Issue #4, opened 2020-05-15
Statuses, Device types, Operation codes etc 


basimsahaf:
1. SCSI statuses 
2. device types
3. operation codes
4. version descriptors (Basic enums)
5. control byte structure (that is shared across many commands).

eyhhuang:
PR: https://github.com/googleinterns/scsi2nvme/pull/16

PR #16, opened 2020-05-15
Basic Enums

eyhhuang:
Implements enums for `Status`, `PeripheralDeviceType`, `OptCode`, and `Version`. Also creates a bitset class for `ControlByte` with getter/setters, ctors, and `getValue()` and adds tests for the class.

Issue: https://github.com/googleinterns/scsi2nvme/issues/4


lib/scsi_defs.h
```diff
@@ -0,0 +1,121 @@
...
+    PERSISTENT_RESERVE_OUT = 0x5f,
+    RESERVE_6 = 0x16,
+    RELEASE_6 = 0x17,
+    START_STOP_UNIT = 0x1b,
+    TEST_UNIT_READY = 0x0,
+  };
+
+  class ControlByte {
+   private:
+    uint8_t obsolete : 2;
```
eyhhuang, line 69: Not sure if uint8_t is the ideal data type for this. The control byte value returned in `getValue()` is only 8 bits, so this seemed to make sense.

test/scsi_defs_test.cc
```diff
@@ -0,0 +1,53 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "gtest/gtest.h"
+#include "lib/scsi_defs.h"
+
+/*
+  Tests the ControlByte class
```
eyhhuang, line 19: I'm not sure if we want all the `scsi_defs` tests to go in this file or if we want separate `.cc` tests for every class. In that case, I'll change the name of this file to `control_byte_test.cc`

lib/scsi_defs.h
```diff
@@ -0,0 +1,121 @@
...
+    uint8_t naca : 1;
+    uint8_t reserved : 3;
+    uint8_t vendorSpecific : 2;
+
+   public:
+    ControlByte() :
+      obsolete{0}, naca{0}, reserved{0}, vendorSpecific{0} {}
+
+    ControlByte(uint8_t obsolete, uint8_t naca, uint8_t reserved, uint8_t vendorSpecific) :
```
harloff-g, line 78: I think it might be better if the constructor arg is simply the control byte and then the constructor parses the control byte to derive obsolete, naca, reserved, and vendorSpecific (or just have both). This is because the raw control byte data is easily accessible to other structures dependent on control byte, but the individual bits are not.

lib/scsi_defs.h
eyhhuang: Could also potentially define constants for these bit sizes

lib/scsi_defs.h
harloff-g: Or, if we want to support direct data casting, I think that the class could only store the single byte and then parse for the individual bit components whenever a getter is called?

lib/scsi_defs.h
```diff
@@ -0,0 +1,121 @@
+#ifndef LIB_SCSI_DEFS_H
+#define LIB_SCSI_DEFS_H
+
+namespace scsi_defs {
+
+  enum Status {
```
harloff-g, line 6: Might be worth explicitly enforcing the enums as one byte (enum Status : char {})

lib/scsi_defs.h
eyhhuang: yeah that would work! I wasn't sure whether we wanted 
individual fields -> control byte  or 
control byte -> individual fields

lib/scsi_defs.h
eyhhuang: enforcing the size is a good idea! What are your thoughts on using uint8_t instead of char

harloff-g: Good point, I think uint8_t is better in this situation

Ralith, line 6: Could these enums all be `enum class`? Such enums, called "strongly typed" or "scoped" enums, tend to be less error-prone as they don't have implicit conversions to/from integers (though `static_cast` can be used for explicit conversion) and they clutter the namespace less.

lib/scsi_defs.h
```diff
@@ -0,0 +1,121 @@
...
+    DO_PREVENT_ALLOW_MEDIUM_REMOVAL = 0x1e,
+    PERSISTENT_RESERVE_IN = 0x5e,
+    PERSISTENT_RESERVE_OUT = 0x5f,
+    RESERVE_6 = 0x16,
+    RELEASE_6 = 0x17,
+    START_STOP_UNIT = 0x1b,
+    TEST_UNIT_READY = 0x0,
+  };
+
+  class ControlByte {
```
Ralith, line 67: Prefer `struct`s with public members for objects whose members would otherwise all be exposed via accessors, and which have no significant built-in logic or invariants.

lib/scsi_defs.h
Ralith: I think consuming a single byte in the constructor makes sense, as the primary use of this code will be to emit/consume said byte.

test/scsi_defs_test.cc
```diff
@@ -0,0 +1,55 @@
...
+    uint8_t obsolete = 0b00;
+    uint8_t naca = 0b1;
+    uint8_t reserved = 0b000;
+    uint8_t vendorSpecific = 0b11;
+    uint8_t expected = 0b11000100;
+
+    scsi_defs::ControlByte *cb = new scsi_defs::ControlByte(obsolete, naca, reserved, vendorSpecific);
+    uint8_t actual = cb->getValue();
+
+    EXPECT_EQ(expected, actual);
```
Ralith, line 34: nit: I'd inline the values of expected/actual directly into this statement (and similarly below) for better clarity/concision.

lib/scsi_defs.h
```diff
@@ -0,0 +1,121 @@
...
+      obsolete{0}, naca{0}, reserved{0}, vendorSpecific{0} {}
+
+    ControlByte(uint8_t obsolete, uint8_t naca, uint8_t reserved, uint8_t vendorSpecific) : 
+      obsolete{obsolete}, naca{naca}, reserved{reserved}, vendorSpecific{vendorSpecific} {}
+
+    uint8_t getValue() const {
+      return 0 | obsolete | (naca << 2) | (reserved << 3) | (vendorSpecific << 6);
+    }
+
+    // Getters
```
w-michal, line 85: I would skip the getters and setters ctor etc and keep it simple - See style recommendation for "passive objects"
https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes

lib/scsi_defs.h
w-michal, line 6: Use enum class. https://abseil.io/tips/86 
Also, the advice in the style says to use kGood, kCheckCondition etc style instead of ALL_CAPS.

eyhhuang: yep that's a good idea!

lib/scsi_defs.h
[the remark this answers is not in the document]
eyhhuang: that's a good idea!

lib/scsi_defs.h
```diff
@@ -0,0 +1,121 @@
...
+    uint8_t vendorSpecific : 2;
+
+   public:
+    ControlByte() :
+      obsolete{0}, naca{0}, reserved{0}, vendorSpecific{0} {}
+
+    ControlByte(uint8_t obsolete, uint8_t naca, uint8_t reserved, uint8_t vendorSpecific) : 
+      obsolete{obsolete}, naca{naca}, reserved{reserved}, vendorSpecific{vendorSpecific} {}
+
+    uint8_t getValue() const {
```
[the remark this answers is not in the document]
eyhhuang: Would we still want a wrapper for bit_cast so we can more easily get the value of the ControlByte? I'm mainly thinking that it would be useful for testing to have a function that gets the value of the control byte

lib/scsi_defs.h
```diff
@@ -0,0 +1,93 @@
+#ifndef LIB_SCSI_DEFS_H
+#define LIB_SCSI_DEFS_H
+
+namespace scsi_defs {
+
+  // SAM-4 Table 33 https://drive.google.com/corp/drive/folders/0B7n7AI-cf5tmN0VnOWd6dnA4RTQ
```
w-michal, line 6: Please don't share internal links.

lib/scsi_defs.h
```diff
@@ -0,0 +1,93 @@
...
+  };
+
+  // SCSI Reference Manual Table 10 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct ControlByte {
+    uint8_t obsolete : 2;
+    uint8_t naca : 1;
+    uint8_t reserved : 3;
+    uint8_t vendorSpecific : 2;
+
+    ControlByte() :
```
w-michal, line 80: Not required - see zero initialization https://en.cppreference.com/w/cpp/language/zero_initialization

lib/scsi_defs.h
```diff
@@ -0,0 +1,93 @@
...
+      obsolete{0}, naca{0}, reserved{0}, vendorSpecific{0} {}
+
+    ControlByte(uint8_t controlByte)  {
```
w-michal, line 83: Not required, 
```
uint_t byte = 0b11000100
ControlByte cb = absl::bit_cast<ControlByte>(byte); 
```
should yield the same result.

lib/scsi_defs.h
```diff
@@ -0,0 +1,93 @@
...
+    kReserve6 = 0x16,
+    kRelease6 = 0x17,
+    kStartStopUnit = 0x1b,
+    kTestUnitReady = 0x0,
+  };
+
+  // SCSI Reference Manual Table 10 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct ControlByte {
+    uint8_t obsolete : 2;
+    uint8_t naca : 1;
```
w-michal, line 76: If the field is 1 bit then it is safe to mark it as bool.

lib/scsi_defs.h
eyhhuang: oops! good catch

lib/scsi_defs.h
```diff
@@ -0,0 +1,84 @@
...
+    kReserve6 = 0x16,
+    kRelease6 = 0x17,
+    kStartStopUnit = 0x1b,
+    kTestUnitReady = 0x0,
+  };
+
+  // SCSI Reference Manual Table 10 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct ControlByte {
+    uint8_t obsolete : 2;
+    bool naca : 1;
```
eyhhuang, line 78: This seems redundant but removing the `: 1` results in issues when trying to bit_cast

lib/scsi_defs.h
```diff
@@ -0,0 +1,84 @@
...
+    kPersistentReserveIn = 0x5e,
+    kPersistentReserveOut = 0x5f,
+    kReserve6 = 0x16,
+    kRelease6 = 0x17,
+    kStartStopUnit = 0x1b,
+    kTestUnitReady = 0x0,
+  };
+
+  // SCSI Reference Manual Table 10 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct ControlByte {
```
basimsahaf, line 76: nit: We can use `typedef` to create a type for ControlByte so that we can declare it in the following way:
```
ControlByte A;
```
instead of 
```
struct ControlByte A;
```

lib/scsi_defs.h
```diff
@@ -0,0 +1,84 @@
...
+    kServiceActionIn = 0x9e,
+    kModeSense6 = 0x1a,
+    kMaintenanceIn = 0xa3,
+    kReadToc = 0x43,
+    kDoPreventAllowMediumRemoval = 0x1e,
+    kPersistentReserveIn = 0x5e,
+    kPersistentReserveOut = 0x5f,
+    kReserve6 = 0x16,
+    kRelease6 = 0x17,
+    kStartStopUnit = 0x1b,
```
w-michal, line 71: Nit - order the enum members by value.

lib/scsi_defs.h
```diff
@@ -0,0 +1,84 @@
...
+    kSpc = 0x3,
+    kSpc2 = 0x4,
+    kSpc3 = 0x5,
+    kSpc4 = 0x6,
+    kSpc5 = 0x7,
+  };
+
+  // SCSI Reference Manual https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  // Opt codes defined for SCSI commands supported by this project; this list may increase
+  enum class OptCode : uint8_t {
```
toadchild, line 57: Should be "opcodes", not "optcodes".  However, the scsi spec seems to prefer "commands".

lib/scsi_defs.h
w-michal: Yes, you have to keep `:1`.

test/scsi_defs_test.cc
```diff
@@ -0,0 +1,35 @@
...
+#include "absl/base/casts.h"
+namespace {
+
+  /*
+    Tests the ControlByte class
+  */
+
+  TEST(ControlByteClass, ShouldReturnCorrectControlByte) {
+    uint8_t cbValue = 0b11000100;
+    scsi_defs::ControlByte cb = absl::lts_2020_02_25::bit_cast<scsi_defs::ControlByte>(cbValue);
```
w-michal, line 27: this looks weird - can't we have just absl::bit_cast?

lib/scsi_defs.h
toadchild: It's not redundant because while a bool is notionally only a single bit of information, it's most likely stored as an entire byte for alignment purposes.

test/scsi_defs_test.cc
eyhhuang: I had this before because I thought it was a compilation issue but it's not needed! I'll make the change

lib/scsi_defs.h
eyhhuang: thanks for catching the "opcode"! Since this enum labels the operation code of each command and a "command" has other data fields in addition to operation code, I feel like it might make more sense to leave this as named OpCode. That being said, I'd be fine changing it to "commands" if that satisfies the specs better.

toadchild: Looks good to me like this.

lib/scsi_defs.h
toadchild: For what it's worth, C++ allows you to use `Foo` as a type if you defined `struct Foo {...}`.  What Basim said is true of C.

basimsahaf: Oh! TIL

PR #19, opened 2020-05-18
Lun structs

### LINKED ISSUE #6: LUN Structures
Structures related to LUN addressing (Lun Data, lun addressing) and report luns command structure.

harloff-g:
closes #6


lib/scsi_defs.h
```diff
@@ -80,5 +88,60 @@ namespace scsi_defs {
...
+    SelectReport selectReport;
+    uint8_t reserved2[3];
+    uint32_t allocLength;
+    uint8_t reserved3[1];
+    ControlByte controlByte;
+  } ABSL_ATTRIBUTE_PACKED;
+
+  // SCSI Reference Manual Table 149 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct ReportLunsData {
+    std::vector<LunAddress> lunList;
```
harloff-g, line 111: Not sure how to allow for an arbitrarily-sized LUN list while also keeping the code lean. Any thoughts would be appreciated.

lib/scsi_defs.h
```diff
@@ -80,5 +88,60 @@ namespace scsi_defs {
...
+
+  // SCSI Reference Manual Table 147 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct ReportLunsCommand {
+    OptCode optCode = OptCode::kReportLuns;
+    uint8_t reserved1[1];
+    SelectReport selectReport;
+    uint8_t reserved2[3];
+    uint32_t allocLength;
+    uint8_t reserved3[1];
+    ControlByte controlByte;
```
basimsahaf, line 106: So far we've been explicitly declaring the number of bits each field is using. You can refer to #16

lib/scsi_defs.h
```diff
@@ -80,5 +88,60 @@ namespace scsi_defs {
...
 
+  // SCSI Reference Manual Table 148 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  enum class SelectReport : uint8_t {
+    kRestrictedMethods = 0x0,
+    kWellKnown = 0x1,
+    kAllLogical = 0x2
+  };
+
+  // SCSI Reference Manual Table 147 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct ReportLunsCommand {
```
basimsahaf, line 99: nit: How about using `typedef` to declare a type for ReportLunsCommand? Refer to https://github.com/googleinterns/scsi2nvme/pull/16#discussion_r426753988

lib/scsi_defs.h
basimsahaf: I have the same question. Would love to hear what others think

lib/scsi_defs.h
```diff
@@ -1,10 +1,18 @@
...
 
+#include <vector>
 #include <cstdint>
 
+#include "absl/base/attributes.h"
+
 namespace scsi_defs {
 
+  const uint32_t LUN_DATA_HEADER_SIZE = 8;
+  const uint32_t LUN_BYTE_SIZE = 8;
```
basimsahaf, line 12: We can use a generic name for this since this will be re-usable in other structures in this file as well

harloff-g: Could you elaborate a bit on this?

basimsahaf: I'm using the same variable [here](https://github.com/googleinterns/scsi2nvme/pull/20/files#diff-74826f7cf214dcd85933fc081f197fbcR10) So, we can just rename it to something generic and not `LUN` specific

harloff-g: Gotcha. I'm primarily using this variable to track the number of bytes in a LUN address. To allow for more dynamic code I'll remove this constant and use sizeof(LunAddress) instead.

lib/scsi_defs.h
```diff
@@ -80,5 +87,60 @@ namespace scsi_defs {
...
 
+  // SCSI Reference Manual Table 148 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  enum class SelectReport : uint8_t {
+    kRestrictedMethods = 0x0,
+    kWellKnown = 0x1,
+    kAllLogical = 0x2
+  };
+
+  // SCSI Reference Manual Table 147 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  typedef struct ReportLunsCommand {
```
eyhhuang, line 98: Do we want a constructor for this? I briefly mentioned this in the chat that having an explicit ctor might help with readability and ease of use later on. Also the ctor can check whether the data is actually providing the correct opcode. Open to thoughts!

lib/scsi_defs.h
```diff
@@ -80,5 +87,60 @@ namespace scsi_defs {
...
+
+  // SCSI Reference Manual Table 149 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  typedef struct ReportLunsData {
+    std::vector<LunAddress> lunList;
+
+    // Form arbitrarily sized vector of LUN lists.
+    ReportLunsData(uint8_t* byteData) {
+      uint8_t listLen = *((uint32_t*)byteData)/sizeof(LunAddress);
+      lunList.reserve(listLen);
+      for (uint32_t i = 0; i < listLen; i++) {
```
eyhhuang, line 116: nit: style guide prefers pre-increment for iterators https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

test/scsi_defs_test.cc
```diff
@@ -16,6 +16,10 @@
 
 #include "gtest/gtest.h"
 #include "absl/base/casts.h"
+
+#include <stdio.h>
+#include <iostream>
```
eyhhuang, line 21: are these two includes used?

test/scsi_defs_test.cc
```diff
@@ -32,4 +36,56 @@ namespace {
     EXPECT_EQ(0b11, cb.vendorSpecific);
   }
 
+  /*
+    Tests the ReportLunsCommand struct
+  */
+
+  TEST(ReportLunsCommandStruct, ShouldBePacked) {
```
eyhhuang, line 43: this is a good idea!

test/scsi_defs_test.cc
harloff-g: Nice catch - they're not

lib/scsi_defs.h
harloff-g: Good point - if we are going to try using constructors, all structs should have one as a standard.

lib/scsi_defs.h
Ralith: I'd favor defining plain helper functions to parse and/or serialize (as needed) the sequence to/from bytes, rather than writing a type that, unlike all the others, isn't POD and can't be bitcast.

lib/scsi_defs.h
```diff
@@ -80,5 +87,60 @@ namespace scsi_defs {
...
+    uint8_t reserved3 : 8;
+    ControlByte controlByte;
+  } ABSL_ATTRIBUTE_PACKED; static_assert(sizeof(ReportLunsCommand) == 12);
+
+  // SCSI Reference Manual Table 149 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct ReportLunsData {
+    std::vector<LunAddress> lunList;
+
+    // Form arbitrarily sized vector of LUN lists.
+    ReportLunsData(uint8_t* byteData) {
```
Ralith, line 113: This should be an `absl::Span<const uint8_t>`. Additionaly, variable names (here and below) [should be snake_case](https://google.github.io/styleguide/cppguide.html#Variable_Names).

lib/scsi_defs.h
```diff
@@ -80,5 +87,60 @@ namespace scsi_defs {
...
+      uint8_t listLen = *((uint32_t*)byteData)/sizeof(LunAddress);
```
Ralith, line 114: This invokes undefined behavior if `byteData` isn't 4-byte-aligned.

lib/scsi_defs.h
```diff
@@ -80,5 +87,60 @@ namespace scsi_defs {
...
+      lunList.reserve(listLen);
+      for (uint32_t i = 0; i < listLen; ++i) {
+        uint32_t byteDataIndex = i * sizeof(LunAddress) + LUN_DATA_HEADER_SIZE;
+        LunAddress tmpLun = *(LunAddress*)(byteData + byteDataIndex);
```
Ralith, line 118: This invokes undefined behavior if `byteData + byteDataIndex` isn't 8-byte-aligned.

lib/scsi_defs.h
```diff
@@ -80,5 +87,60 @@ namespace scsi_defs {
...
+
+  // SCSI Reference Manual Table 149 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct ReportLunsData {
+    std::vector<LunAddress> lunList;
+
+    // Form arbitrarily sized vector of LUN lists.
+    ReportLunsData(uint8_t* byteData) {
+      uint8_t listLen = *((uint32_t*)byteData)/sizeof(LunAddress);
+      lunList.reserve(listLen);
+      for (uint32_t i = 0; i < listLen; ++i) {
```
Ralith, line 116: Before reading from memory based on a passed-in length, we should check that the length does not exceed the size of the passed-in memory region, and return an error if it does. Since constructors cannot return errors, this constructor will need to be converted into a function instead.

lib/scsi_defs.h
```diff
@@ -80,5 +87,60 @@ namespace scsi_defs {
...
+        uint32_t byteDataIndex = i * sizeof(LunAddress) + LUN_DATA_HEADER_SIZE;
+        LunAddress tmpLun = *(LunAddress*)(byteData + byteDataIndex);
+        lunList.push_back(tmpLun);
+      }
+    }
+
+    // Returns a vector of bytes describing the SCSI data.
+    std::vector<uint8_t> getValue() {
```
Ralith, line 124: Accessors [should be named like variables](https://google.github.io/styleguide/cppguide.html#Function_Names), i.e. snake_case. Also consider allowing the caller to pass in memory to write to, rather than allocating memory internally.

lib/scsi_defs.h
toadchild: The standard approach is to have a struct for the header and a struct for the list elements, and then to have the code know how to glue them together.

lib/scsi_defs.h
```diff
@@ -3,8 +3,12 @@
 
 #include <cstdint>
 
+#include "absl/base/attributes.h"
+
 namespace scsi_defs {
 
+  typedef uint64_t LunAddress;
```
eyhhuang, line 10: is this used somewhere

harloff-g: It's not currently used as LunAddresses are reported as variable sized data. The intention was for the typedef to be used later when variable sized fields are handled.

eyhhuang: ah I see. thanks for clarifying!

lib/scsi_defs.h
```diff
@@ -8,6 +8,8 @@
 // The fields of structures in this namespace are arranged according to Big Endian format.
 namespace scsi_defs {
 
+  typedef uint64_t LunAddress;
```
Ralith, line 11: nit: [Prefer `using`](https://google.github.io/styleguide/cppguide.html#Aliases).

lib/scsi_defs.h
```diff
@@ -259,5 +261,32 @@ namespace scsi_defs {
...
+    uint32_t alloc_length : 32;
+    uint8_t reserved_3 : 8;
+    ControlByte control_byte;
+  } ABSL_ATTRIBUTE_PACKED;
+  static_assert(sizeof(ReportLunsCommand) == 12);
+
+  // SCSI Reference Manual Table 149 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  // This struct is a header for variable sized data
+  struct ReportLunsParamData {
+    uint32_t list_byte_length : 32;
```
toadchild, line 286: Bitfield annotations aren't necessary if you're using native sizes (8, 16, 32, 64 bits).

harloff-g: Basim had found this stack overflow post that seems to indicate that the bitfields do cause some obscure difference where the size of the structure can be incorrectly packed if they're not used. Emily was able to replicate the discrepancy in our code base using the example in the post. Thoughts on this?
https://stackoverflow.com/questions/25822679/packed-bit-fields-in-c-structures-gcc

toadchild: Nothing wrong with consistency.  Comment withdrawn.

Issue #7, opened 2020-05-15
Inquiry Command Structures


basimsahaf:
Inquiry command structures (with supporting structures and enums) we want to support inquiring: 
1. standard data
2. Device ID
3. Unit serial number
4. Block Limits and device characteristics
5. logical block provisioning pages.

PR #20, opened 2020-05-18
Inquiry Command Structure 

basimsahaf:
Code for the Inquiry Command Structure. The current base branch is `sense-enums`. Will rebase once `sense-enums` is merged to master.

This is one part of the #7 issue and only contains `InquiryCommand` structure. Will be putting up another PR for `InquiryCommandData`


lib/scsi_defs.h
```diff
@@ -108,5 +115,29 @@ namespace scsi_defs {
...
+  // Refer to https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf , Section 3.6 Table 58
+  typedef struct InquiryCommand {
+    OptCode optCode = OptCode::kInquiry;
+    bool evpd: 1; // Enable Vital Product Data (EVPD)
+    bool obsolete: 1; // formerly CMDDT
+    uint8_t reserved: 6;
+    uint8_t pageCode: 8;
+    uint16_t allocationLength: 16;
+    uint8_t control: 8;
+    uint16_t padding: 16;
```
harloff-g, line 128: This seems hacky imo. As far as I can tell, there's nothing in the stl standard that guarantees the internal size of a bitset object. If we want to use absl bit casting, I think using a structure with guaranteed size makes more sense.

lib/scsi_defs.h
```diff
@@ -108,5 +115,29 @@ namespace scsi_defs {
     kCompleted = 0xF
   };
 
+
+  // Refer to https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf , Section 3.6 Table 58
+  typedef struct InquiryCommand {
```
Ralith, line 120: `typedef`ing structs isn't needed in C++

basimsahaf: changed to `struct`

test/scsi_defs_test.cc
```diff
@@ -32,4 +34,72 @@ namespace {
...
+    (evpd << 8) +
+    (obsolete << 9) +
+    (reserved << 10) +
+    (pageCode << 16) + 
+    (allocLength << 24) +
+    (control << 40));
+
+    // 2. using bit_cast:
+
+    std::optional<scsi_defs::InquiryCommand> ic = scsi_defs::InquiryCommand::fromBits(cmdValue);
```
basimsahaf, line 92: We're constructing the command objects using bit_cast here.

lib/scsi_defs.h
basimsahaf: I agree! I made a new PR to review the memcpy way of constructing the object #24 . Let me know your thoughts on that and we can use that instead

test/scsi_defs_test.cc
Ralith: I think `bit_cast` is a good approach to use in unit tests, but it would be better to cast to/from `std::array<uint8_t, N>` where N is the size of the structure as defined in the spec.

lib/scsi_defs.h
```diff
@@ -3,8 +3,14 @@
 
 #include <cstdint>
 
+#include "absl/base/attributes.h"
+#include "absl/base/casts.h"
+
+
 namespace scsi_defs {
 
+  const uint32_t kInquiryCommandSize = 6;
```
Ralith, line 12: Prefer `constexpr` for POD global constants like this.

lib/scsi_defs.h
```diff
@@ -3,8 +3,14 @@
 
 #include <cstdint>
 
+#include "absl/base/attributes.h"
+#include "absl/base/casts.h"
```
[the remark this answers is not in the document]
basimsahaf: Good catch. I updated the code and removed the header

lib/scsi_defs.h
```diff
@@ -108,5 +114,18 @@ namespace scsi_defs {
...
+  // Refer to https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf , Section 3.6 Table 58
+  // The fields are arrange according to Big Endian format.
+  struct InquiryCommand {
+    OpCode opCode = OpCode::kInquiry;
+    uint8_t reserved: 6;
+    bool obsolete: 1; // formerly CMDDT
+    bool evpd: 1; // Enable Vital Product Data (EVPD)
+    uint8_t pageCode: 8;
+    uint16_t allocationLength: 16;
+    ControlByte controlByte;
```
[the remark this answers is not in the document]
basimsahaf: changed the case in this PR. Will make a new ticket to update the names elsewhere as well

lib/scsi_defs.h
basimsahaf: Good suggestion. I updated the code to use `constexpr`

lib/scsi_defs.h
```diff
@@ -108,5 +112,17 @@ namespace scsi_defs {
     kCompleted = 0xF
   };
 
+  // Refer to https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf , Section 3.6 Table 58
+  // The fields are arrange according to Big Endian format.
```
eyhhuang, line 116: nit: arranged instead of arrange

lib/scsi_defs.h
```diff
@@ -2,9 +2,13 @@
 #define LIB_SCSI_DEFS_H
 
 #include <cstdint>
+#include "absl/base/attributes.h"
```
eyhhuang, line 5: nit: line break between this and the previous include

lib/scsi_defs.h
```diff
@@ -2,9 +2,13 @@
...
+
 
 namespace scsi_defs {
 
+  constexpr uint32_t kInquiryCommandSize = 6;
```
eyhhuang, line 10: no longer doing these constants https://github.com/googleinterns/scsi2nvme/pull/29#discussion_r428935909

lib/scsi_defs.h
```diff
@@ -108,5 +112,18 @@ namespace scsi_defs {
     kCompleted = 0xF
   };
 
+  // Refer to https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf , Section 3.6 Table 58
+  // The fields are arranged according to Big Endian format.
```
eyhhuang, line 88: Since we decided to just have one comment about Big Endian, can we move this comment to the top of all structs or something?

lib/scsi_defs.h
basimsahaf: I thought it was only for the read, write, sync and verify structs because of their naming. Let me link this discussion on #29 and have folks pitch in

lib/scsi_defs.h
basimsahaf: Yes, we can put this at the top of the file. Making the change right now.

lib/scsi_defs.h
Ralith: Responded there, but tl;dr they don't seem necessary.

Issue #8, opened 2020-05-15
 Unit ready command structures


basimsahaf:
"Unit ready" command structures.

PR #21, opened 2020-05-18
Add TestUnitReady command

eyhhuang:
Implements TestUnitReady command. 

Issue: https://github.com/googleinterns/scsi2nvme/issues/8


lib/scsi_defs.h
```diff
@@ -80,5 +85,23 @@ namespace scsi_defs {
     uint8_t vendorSpecific : 2;
   } ControlByte;
 
+  //SCSI Reference Manual Table 202 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  typedef struct TestUnitReadyCommand {
+    OpCode opCode = OpCode::kTestUnitReady;
+    uint32_t reserved;
+    ControlByte controlByte;
+    uint16_t padding;   // Necessary for bit_cast
```
eyhhuang, line 93: So the `TestUnitReadyCommand` as per the specs is 48 bits (6 bytes). But when I make a `bitset<48>`, the size of that is actually 8 bytes instead of 6 bytes. So to allow for `bit_cast` which requires the `sizeof()` to be the same, I padded my struct with 2 bytes so it can be 8 bytes.

lib/scsi_defs.h
```diff
@@ -80,5 +85,23 @@ namespace scsi_defs {
...
+
+    TestUnitReadyCommand() {}
+
+    TestUnitReadyCommand(std::bitset<48> cmd) {
+      if (OpCode::kTestUnitReady != (OpCode) (cmd.to_ulong() & 8)) {
+        throw std::runtime_error("Invalid OpCode for Test Unit Ready Command");
```
harloff-g, line 99: I'm wondering if it would be better to use a custom exception class(es) or not

test/scsi_defs_test.cc
```diff
@@ -32,4 +36,25 @@ namespace {
...
+    scsi_defs::TestUnitReadyCommand command;
+    EXPECT_EQ(8, sizeof(command));
+  }
+
+  TEST(TestUnitReadyCommandStruct, ShouldReturnCorrectTestUnitReadyCommand) {
+    scsi_defs::OpCode opCode = scsi_defs::OpCode::kTestUnitReady;
+    uint64_t reserved = 8927;
+    uint64_t controlByte = 12;
+
+    std::bitset<48> cmdValue(static_cast<uint8_t>(opCode) + (reserved << 8) + (controlByte << 40));
```
harloff-g, line 52: I believe that SCSI is big endian, and this seems to be assuming little endian (https://www.t10.org/ftp/t10/document.03/03-127r0.pdf). I'm not 100% on this so definitely let me know if I'm wrong!

lib/scsi_defs.h
eyhhuang: Yeah we could make an InvalidOpCodeException (or a different name)

lib/scsi_defs.h
```diff
@@ -0,0 +1,107 @@
...
+    OpCode opCode = OpCode::kTestUnitReady;
+    uint32_t reserved;
+    ControlByte controlByte;
+    uint16_t padding;   // Necessary for bit_cast
+
+    TestUnitReadyCommand() {}
+
+    TestUnitReadyCommand(std::bitset<48> cmd) {
+      if (OpCode::kTestUnitReady != (OpCode) (cmd.to_ulong() & 0b11111111)) {
+        throw std::runtime_error("Invalid OpCode for Test Unit Ready Command");
```
eyhhuang, line 99: TODO figure out another way to handle this error instead of throwing exceptions

lib/scsi_defs.h
Ralith: What's the benefit of caring about `bitset` here? Ultimately we'll be wanting to convert to/from buffers of raw bytes containing the specified ABI, so we should try to respect that ABI exactly. In other words, if the spec says a structure should be 6 bytes, the definition should probably be 6 bytes.

toadchild: I'm concerned that changing the size will lead to problems when we're looking at actual scsi command data - those aren't going to include padding bytes for us.  If bitset isn't working, we might have to fall back on more old fashioned methods such as C-style byte arrays.

lib/scsi_defs.h
```diff
@@ -80,5 +85,21 @@ namespace scsi_defs {
     uint8_t vendorSpecific : 2;
   } ControlByte;
 
+  //SCSI Reference Manual Table 202 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  typedef struct TestUnitReadyCommand {
```
Ralith, line 89: The `typedef struct` pattern isn't needed in C++; you can just declare the struct the same way you would a class.

lib/scsi_defs.h
Ralith: `std::array<uint8_t, 6>` would also be a nice robust approach.

lib/scsi_defs.h
```diff
@@ -80,5 +85,21 @@ namespace scsi_defs {
...
+    OpCode opCode = OpCode::kTestUnitReady;
+    uint32_t reserved;
+    ControlByte controlByte;
+    uint16_t padding;   // Necessary for bit_cast
+
+    static std::optional<TestUnitReadyCommand> fromBits(std::bitset<48> cmd) {
```
w-michal, line 95: I am working on next steps, and one would be parser and handler of raw data. I would prefer to have all parsing and handling in a single place instead of having it across all the structures.

eyhhuang: So instead of having "constructors" for each class, we handle that logic elsewhere?

test/scsi_defs_test.cc
```diff
@@ -32,4 +35,12 @@ namespace {
     EXPECT_EQ(0b11, cb.vendorSpecific);
   }
 
+  /*
+    Tests the TestUnitReadyCommand sruct
+  */
+  TEST(TestUnitReadyCommandStruct, ShouldBePacked) {
```
harloff-g, line 41: We can replace this test with a static assert (see https://github.com/googleinterns/scsi2nvme/pull/24#discussion_r427580382)

lib/scsi_defs.h
```diff
@@ -1,10 +1,13 @@
 #ifndef LIB_SCSI_DEFS_H
 #define LIB_SCSI_DEFS_H
 
-#include <cstdint>
+#include "absl/base/attributes.h"
+#include "absl/base/casts.h"
```
harloff-g, line 5: Is this include used other than for uint_t's? #include <cstdint> might be a more concise include if that's the case.

lib/scsi_defs.h
```diff
@@ -73,12 +76,19 @@ namespace scsi_defs {
   };
 
   // SCSI Reference Manual Table 10 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
-  typedef struct ControlByte {
+  struct ControlByte {
     uint8_t obsolete : 2;
     bool naca : 1;
     uint8_t reserved : 3;
     uint8_t vendorSpecific : 2;
```
harloff-g, line 83: Can this PR also change this variable to snake_case? This would close all style issues in master

eyhhuang: done!

lib/scsi_defs.h
```diff
@@ -2,7 +2,7 @@
 #define LIB_SCSI_DEFS_H
 
 #include "absl/base/attributes.h"
-#include "absl/base/casts.h"
+#include "cstdint"
```
Ralith, line 5:
```suggestion
#include <cstdint>
```

Always use angle brackets for system headers.

lib/scsi_defs.h
```diff
@@ -1,10 +1,13 @@
 #ifndef LIB_SCSI_DEFS_H
 #define LIB_SCSI_DEFS_H
 
+#include "absl/base/attributes.h"
 #include <cstdint>
```
Ralith, line 5: Standard library headers [should come first and be separated by an empty line](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes).

eyhhuang: Good catch! I'll make sure to be more careful with the includes next time!

lib/scsi_defs.h
```diff
@@ -82,6 +82,32 @@ namespace scsi_defs {
     uint8_t vendor_specific : 2;
   };
 
+  // SCSI Reference Manual Table 202 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct TestUnitReadyCommand {
+    OpCode op_code = OpCode::kTestUnitReady;
+    uint32_t reserved : 32;;
```
toadchild, line 88: Bitfield annotations aren't necessary if you're using native sizes (8, 16, 32, 64 bits).

eyhhuang: We actually had a discussion about this on this thread https://github.com/googleinterns/scsi2nvme/pull/28#pullrequestreview-418531627
This was partly for style reasons

lib/scsi_defs.h
```diff
@@ -82,6 +82,32 @@ namespace scsi_defs {
     uint8_t vendor_specific : 2;
   };
 
+  // SCSI Reference Manual Table 202 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct TestUnitReadyCommand {
+    OpCode op_code = OpCode::kTestUnitReady;
```
Ralith, line 87: Should we split these out? [Discussion elsewhere](https://github.com/googleinterns/scsi2nvme/pull/30#discussion_r432001349).

lib/scsi_defs.h
Ralith, line 88:
```suggestion
    uint32_t reserved : 32;
```
Stray semicolon.

eyhhuang: great catch!

toadchild: Ok, sounds good.

PR #23, opened 2020-05-19
Draft CI actions workflow

### LINKED ISSUE #18: Continuous integration
Set up automated build and test of every pull request and commit to master. Helps prevent inadvertent breakage from slipping in.

There are a number of free CI services available for open source projects, but [GitHub Actions](https://help.github.com/en/actions) predictably have the best integration. Testing on Linux only should be fine.

harloff-g:
Closes #18 

Adds a Github actions workflow which Bazel builds & tests.
Workflow runs on all pull requests, and any pushes to master.


.github/workflows/main.yml
```diff
@@ -0,0 +1,40 @@
...
+# events but only for the master branch. Triggers on all pull requests.
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
```
Ralith, line 15: "build" might not be the most apt name, since this also runs tests.

harloff-g: Good point. I updated the name to holistically describe the check.

.github/workflows/main.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Install bazelisk
+      run: |
+        curl -LO "https://github.com/bazelbuild/bazelisk/releases/download/v1.4.0/bazelisk-linux-amd64"
+        mkdir -p "${GITHUB_WORKSPACE}/bin/"
+        mv bazelisk-linux-amd64 "${GITHUB_WORKSPACE}/bin/bazel"
+        chmod +x "${GITHUB_WORKSPACE}/bin/bazel"
+    
+    - name: Use c++ 17
+      run: |
+        export BAZEL_CXXOPTS="-std=c++17"
```
Ralith, line 36: Rather than an additional imperative step, could this be set using [the declarative environment variable support](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables)?

.github/workflows/main.yml
```diff
@@ -0,0 +1,42 @@
...
+        curl -LO "https://github.com/bazelbuild/bazelisk/releases/download/v1.4.0/bazelisk-linux-amd64"
+        mkdir -p "${GITHUB_WORKSPACE}/bin/"
+        mv bazelisk-linux-amd64 "${GITHUB_WORKSPACE}/bin/bazel"
+        chmod +x "${GITHUB_WORKSPACE}/bin/bazel"
+        
+    - name: Build
+      run: |
+        "${GITHUB_WORKSPACE}/bin/bazel" build ...
+      env:
+        BAZEL_CXXOPTS: -std=c++17
```
Ralith, line 38: Are you certain this only needs to be set for the build stage? In other words, is there any risk that the test stage might see the different flags and decide it needs to rebuild things for consistency? If there might be such a risk, it would be safer to set this for the entire job, or even the entire file (it looks like `env` works [as a top-level key](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#env)).

harloff-g: After testing locally, it looks like this can cause build errors during Bazel test. I've updated the workflow to bind the environment variable at the job level

.github/workflows/main.yml
```diff
@@ -0,0 +1,42 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build-and-test:
+    env:
+        BAZEL_CXXOPTS: -std=c++17
```
Ralith, line 17: Excess indentation?

Issue #11, opened 2020-05-15
Read Capacity


basimsahaf:
Read capacity command structures.

PR #25, opened 2020-05-20
Read capacity

eyhhuang:
Implements ReadCapacity10 Command and Data. 
ReadCapacity16 Command and Data was _not_ implemented because it is not currently one of the commands to be supported.

Issue: https://github.com/googleinterns/scsi2nvme/issues/11


lib/scsi_defs.h
```diff
@@ -90,5 +92,21 @@ namespace scsi_defs {
...
+    uint32_t logical_block_address; // obsolete Big endian
+    uint16_t reserved_2;
+    uint8_t reserved_3; // obsolete PMI bit
+    ControlByte control_byte;
+  } ABSL_ATTRIBUTE_PACKED; static_assert(sizeof(ReadCapacity10Command) == kReadCapacity10CommandSize);
+
+  // SCSI Reference Manual Table 120 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct ReadCapacity10Data {
+    uint32_t returned_logical_block_address; // Big endian
+    uint32_t block_length; //Big endian
```
Ralith, line 108:
```suggestion
    uint32_t block_length; // Big endian
```

Also, do we need these comments on every field? If all multibyte fields are big-endian it would be clearer to just document that up front.

PR #26, opened 2020-05-20
InquiryData structure

### LINKED ISSUE #7: Inquiry Command Structures

basimsahaf:
Closes #7


lib/scsi_defs.h
```diff
@@ -125,5 +126,65 @@ namespace scsi_defs {
...
+    bool ius: 1; // Information Units Supported bit
+    uint8_t reserved_10: 8;
+    uint16_t vendor_descriptor_1;
+    uint16_t vendor_descriptor_2;
+    uint16_t vendor_descriptor_3;
+    uint16_t vendor_descriptor_4;
+    uint16_t vendor_descriptor_5;
+    uint16_t vendor_descriptor_6;
+    uint16_t vendor_descriptor_7;
+    uint16_t vendor_descriptor_8;
```
basimsahaf, line 183: I'm wondering if we can merge these fields into one single field. Can have an array of elements of 16 bits each

lib/scsi_defs.h
```diff
@@ -125,5 +126,63 @@ namespace scsi_defs {
...
+    bool ius: 1; // Information Units Supported bit
+    uint8_t reserved_8: 8;
+    uint16_t vendor_descriptor_1;
+    uint16_t vendor_descriptor_2;
+    uint16_t vendor_descriptor_3;
+    uint16_t vendor_descriptor_4;
+    uint16_t vendor_descriptor_5;
+    uint16_t vendor_descriptor_6;
+    uint16_t vendor_descriptor_7;
+    uint16_t vendor_descriptor_8;
```
basimsahaf, line 181: I am wondering if we can merge all these fields into one single field. So we can have an array of elements of 16-bit each

lib/scsi_defs.h
```diff
@@ -125,5 +126,63 @@ namespace scsi_defs {
...
+    uint8_t reserved_9[22];
+    std::vector<uint8_t> vendor_specific; // this field has a variable size
```
basimsahaf, line 183: This field is of variable size. So, I made it a vector so that whenever this struct is initialized, this vector is built in the constructor. Let me know if you think there's a better way to do this

basimsahaf: I just realized that adding a vector is an unpacked non-POD field. This will make the compiler ignore the packed attribute.

lib/scsi_defs.h
Ralith: That should be fine, yeah--and indeed much more convenient.

lib/scsi_defs.h
Ralith: I think the way to go here is omit this field from this structure, and then in future parser logic take care to handle it specially.

basimsahaf: Sounds good! Will omit this field for now

lib/scsi_defs.h
```diff
@@ -97,5 +99,77 @@ namespace scsi_defs {
...
+
+  // Refer to https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf , Section 3.6.2 Table 63
+  enum class TPGS : uint8_t {
+    kNotSupported = 0b00,
+    kImplicitAccess = 0b01,
+    kExcplicitAccess = 0b10,
+    kFullAccess = 0b11,
+  };
+
+  enum class ResponseDataFormat : uint8_t {
```
eyhhuang, line 119: Can we add a reference to this?

lib/scsi_defs.h
```diff
@@ -97,5 +99,77 @@ namespace scsi_defs {
...
+    bool reserved_5: 1;
+    bool wbus_16: 1; // Wide Bus bit 
+    bool sync: 1;
+    bool obsolete_4: 1;
+    bool reserved_6: 1;
+    bool cmdque: 1; // Command Management Model bit
+    bool vs_2: 1; // vendor specific bit
+    uint8_t vendor_identification[8];
+    uint8_t product_identification[16];
+    uint8_t product_revision_level[4];
```
eyhhuang, line 162: So for this field, we can also do `uint32_t` instead of a byte array. Should we set a standard of when we should just a byte array and when we should just use a primitive when both options work?

lib/scsi_defs.h
```diff
@@ -97,5 +99,77 @@ namespace scsi_defs {
...
+    uint8_t vendor_specific_1[20];
+    uint8_t reserved_7: 4;
+    uint8_t clocking: 2;
+    bool qas: 1; // Quick Arbitration and Selection Supported bit
+    bool ius: 1; // Information Units Supported bit
+    uint8_t reserved_8: 8;
+    uint16_t vendor_descriptor[8];
```
eyhhuang, line 169: can we rename this `vendor_descriptors`? I was initially confused by it was an array of `uint16_t` but then I realized it was actually an array of 8 `vendor_descriptor`, each with size `uint16_t`

lib/scsi_defs.h
```diff
@@ -97,5 +99,77 @@ namespace scsi_defs {
...
+    uint8_t reserved_9[22];
+  } ABSL_ATTRIBUTE_PACKED; 
+  static_assert(sizeof(InquiryData) >= kInquiryDataSize);
```
eyhhuang, line 172: should this be `>= 96`.

lib/scsi_defs.h
harloff-g: Additionally, vendor_identification could be declared as a uint64_t

lib/scsi_defs.h
harloff-g: Why are we asserting greater than or equal to and not just equal to?

lib/scsi_defs.h
```diff
@@ -97,5 +99,77 @@ namespace scsi_defs {
...
+  // Refer to https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf , Section 3.6.2 Table 59
+  // Fields are arranged according to Big Endian format
+  struct InquiryData {
+    PeripheralQualifier peripheral_qualifier: 3;
+    PeripheralDeviceType peripheral_device_type: 5;
+    bool rmb: 1; // Removable Media Bit (RMB)
+    bool lu_cong: 1;
+    uint8_t reserved_1: 6;
+    Version version: 8;
+    bool reserved_2: 2;
```
harloff-g, line 134: I think it would be more consistent to reserve bool to boolean fields only (i.e. one bit)

lib/scsi_defs.h
basimsahaf: There's actually no reference to this in the Seagate documentation. Its there on the internal documentation that we have. Not sure how to provide a reference in this case

lib/scsi_defs.h
basimsahaf: That's a good catch! And I agree with having a standard for using an array vs a primitive. I would say that if we have a field > 64 bits, we should use an array and in all other cases, we should use the appropriate primitive (eg uint8_t, uint16_t etc)

lib/scsi_defs.h
basimsahaf: Yup, will make the change. Sorry for the confusion

lib/scsi_defs.h
basimsahaf: That's a typo. Thanks for catching that.

lib/scsi_defs.h
eyhhuang: i think the `>=` is because of the variable size fields

basimsahaf: @eyhhuang the official documentation says that this structure shall contain at least 36 bytes and that's why I am asserting for that value.

@harloff-g This structure had a variable sized field and the size of this structure is at least 36 bytes. That's why I was checking for `>= 36` but since we're not including the variable sized fields as of now, I change it to assert for the exact value.

basimsahaf: I think we will need to check for 96 bytes in the static assert but will need to have a check in the parsing logic later on for ensuring at least 36 bytes are passed to this structure.

lib/scsi_defs.h
```diff
@@ -97,5 +99,77 @@ namespace scsi_defs {
...
+  enum class ResponseDataFormat : uint8_t {
+    kObsolete0 = 0x0,
+    kObsolete1 = 0x1,
+    kCompliant = 0x2,
+  };
+
+  // Refer to https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf , Section 3.6.2 Table 59
+  // Fields are arranged according to Big Endian format
+  struct InquiryData {
+    PeripheralQualifier peripheral_qualifier: 3;
```
harloff-g, line 128: nit: master currently defines bit fields with a space between the colon and the variable i.e. `PeripheralQualifier peripheral_qualifier : 3;`

lib/scsi_defs.h
harloff-g: Yeah that's kind of what I was thinking as well. The struct will always be 96 bytes but some fields might not be valid depending on the value of the additional_length field

lib/scsi_defs.h
basimsahaf: Updated the formatting. Nice catch 👍

lib/scsi_defs.h
```diff
@@ -3,12 +3,15 @@
...
 
+
 #include "absl/base/attributes.h"
 
 // The fields of structures in this namespace are arranged according to Big Endian format.
 namespace scsi_defs {
 
+  // size of structs in bytes
   constexpr uint32_t kInquiryCommandSize = 6;
+  constexpr uint32_t kInquiryDataSize = 96;
```
Ralith, line 14: Seems like we're moving away from defining size constants.

lib/scsi_defs.h
basimsahaf: Yup so I updated the assert to check for 96 here. We can enforce the 36 byte requirement later on

lib/scsi_defs.h
basimsahaf: > Additionally, vendor_identification could be declared as a `uint64_t`

@harloff-g do you mean that we should declare `vendor_identification` as an array of `uint64_t`? @eyhhuang and I discussed that its better to have an array with elements of 1 byte each. What are your views on that?

harloff-g: No, I just mean that vendor_identification doesn't have to be an array at all and can be declared as `uint64_t vendor_identification : 64;`

lib/scsi_defs.h
basimsahaf: Removed constants

lib/scsi_defs.h
basimsahaf: Oh true! Sorry, I actually misread your previous comment to mean `product_identification`. That makes sense and I will update the field 👍

PR #27, opened 2020-05-20
Mode Sense Structs

### LINKED ISSUE #12: Mode Sense Command
"Mode sense" command structures.

harloff-g:
Closes #12 
Adds two POD structs to contain Mode Sense command data.


lib/scsi_defs.h
```diff
@@ -80,5 +83,37 @@ namespace scsi_defs {
     uint8_t vendorSpecific : 2;
   } ControlByte;
 
+  // SCSI Reference Manual Table 73 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  // All fields big-endian
+  struct ModeSense6Command {
+    OpCode opCode = OpCode::kModeSense6;
+    uint8_t reserved_1 : 4;
+    bool dbd : 1;
```
basimsahaf, line 91: nit: having a short comment for non-obvious field names would be helpful in the future. Non-obvious field names include abbreviated names like `dbd`, `pc` etc

lib/scsi_defs.h
```diff
@@ -80,5 +83,37 @@ namespace scsi_defs {
...
+    uint8_t pc : 2;
+    uint8_t page_code : 6;
+    uint8_t sub_page_code : 8;
+    uint8_t alloc_length : 8;
+    ControlByte control_byte;
+  } ABSL_ATTRIBUTE_PACKED;
+  static_assert(sizeof(ModeSense6Command) == 6);
+
+  // SCSI Reference Manual Table 75 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  // All fields big-endian
```
basimsahaf, line 102: I added a comment for this on the top of the namespace in #20. So we can remove this particular comment from here as #20 will take care of it

harloff-g: Good catch, I forgot to remove this

lib/scsi_defs.h
```diff
@@ -80,5 +83,35 @@ namespace scsi_defs {
...
+    uint8_t reserved_1 : 3;
+    bool llbaa : 1; // Long LBA accepted
+    bool dbd : 1; // Disable block descriptors
+    uint8_t reserved_2 : 3;
+    uint8_t pc : 2; // Page control
+    uint8_t page_code : 6;
+    uint8_t sub_page_code : 8;
+    uint32_t reserved_3 : 24;
+    uint16_t alloc_length : 16;
+    ControlByte controlByte;
```
eyhhuang, line 112: nit: this should be `control_byte`

harloff-g: Nice catch!

Issue #13, opened 2020-05-15
Persistent Reservation


basimsahaf:
Persistent reservation command structures. (reserve in, read reservation, reserve out)

PR #28, opened 2020-05-21
Persistent reserve

eyhhuang:
Closes https://github.com/googleinterns/scsi2nvme/issues/13


lib/scsi_defs.h
```diff
@@ -90,5 +95,63 @@ namespace scsi_defs {
     ControlByte control_byte;
   } ABSL_ATTRIBUTE_PACKED; static_assert(sizeof(TestUnitReadyCommand) == kTestUnitReadyCommandSize);
 
+  // SCSI Reference Manual Table 76 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct PersistentReserveInCommand {
+    OpCode op_code = OpCode::kPersistentReserveIn;
+    uint8_t reserved_1 : 3;
+    uint8_t service_action : 5;
+    uint64_t reserved_2 : 40;
+    uint16_t allocation_length; // Big endian
```
Ralith, line 104: Are there any multibyte integers in SCSI that *aren't* big-endian? If not, we probably don't need to annotate individual fields.

eyhhuang: yeah you're right. all of the multi-byte fields are big-endian. There was a comment on an earlier PR about annotating the endianess of fields, but since they're all big-endian, we can jsut have one comment somewhere

lib/scsi_defs.h
```diff
@@ -90,5 +95,63 @@ namespace scsi_defs {
...
 
+  // SCSI Reference Manual Table 76 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct PersistentReserveInCommand {
+    OpCode op_code = OpCode::kPersistentReserveIn;
+    uint8_t reserved_1 : 3;
+    uint8_t service_action : 5;
+    uint64_t reserved_2 : 40;
+    uint16_t allocation_length;
+    ControlByte control_byte;
+  } ABSL_ATTRIBUTE_PACKED; static_assert(sizeof(PersistentReserveInCommand) == kPersistentReserveInCommandSize);
```
w-michal, line 106: prefer having `static_assert` in a separate line.

eyhhuang: done!

lib/scsi_defs.h
```diff
@@ -108,5 +108,68 @@ namespace scsi_defs {
...
+  } ABSL_ATTRIBUTE_PACKED;
+  static_assert(sizeof(PersistentReserveOutCommand) == 10);
+
+  // Persistent Reserve Out Parameter List
+  // Used by Persistent Reserve Out command for any service action other than Register and Move
+  // SCSI Reference Manual Table 90 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct ProParamList {
+    uint64_t reservation_key;
+    uint64_t service_action_reservation_key;
+    uint32_t obsolete_1;
```
basimsahaf, line 162: nit: In some of our previous PRs, we are mentioning the number of bits used by the field even if it is using all bits of the aforementioned type. I suggest having a standard for this.

eyhhuang: I can see how specifying the bits for all fields can introduce clarity but I wonder if it might unnecessarily clutter the code. Since we assert the size of each struct, I also don't think this will add much safety. I'd love to hear more views!

harloff-g: I think Emily is right that doing so doesn't really add meaningful size safety. So, the choice is mostly stylistic. I could go either way.

basimsahaf: Yeah, I agree that there's no upside to it in terms of logic. I am fine going either way too but we should fix this before merging our branches so that our code is consistent everywhere

eyhhuang: i updated my changes so they specify the size!

basimsahaf: Just for future reference: https://stackoverflow.com/questions/25822679/packed-bit-fields-in-c-structures-gcc

PR #29, opened 2020-05-21
Read/Write/Verify/Sync Commands

### LINKED ISSUE #10: Read, Write, Verify and Sync Cache Commands
Read, Write, Verify and synchronize cache commands structures. Please note that some commands exist in a few variants like read6, read10 and read16. (They all act essentially the same, but *10 and *16 include more options)

basimsahaf:
Closes #10


lib/scsi_defs.h
```diff
@@ -2,9 +2,29 @@
...
+  constexpr uint32_t kWrite12CommandSize = 12;
+  constexpr uint32_t kWrite16CommandSize = 16;
+  constexpr uint32_t kWrite32CommandSize = 32;
+  constexpr uint32_t kVerify6CommandSize = 6;
+  constexpr uint32_t kVerify10CommandSize = 10;
+  constexpr uint32_t kVerify12CommandSize = 12;
+  constexpr uint32_t kVerify16CommandSize = 16;
+  constexpr uint32_t kVerify32CommandSize = 32;
+  constexpr uint32_t kSynchronizeCache10Size = 10;
+  constexpr uint32_t kSynchronizeCache16Size = 16;
```
basimsahaf, line 25: I just realized that we can club all of these together since they are all standard byte sizes. Any suggestions on naming?

Ralith: I'm not sure what you mean by "club together". In this particular case where the intended size of the structure is always exactly what's in the name, I'm not sure I'd bother with the variables at all, though others might feel differently.

basimsahaf: What I meant by "club together" is that since all these structures are either 6, 10, 12, 16 or 32 bytes, we can have variables for these sizes which all these structures can re-use instead of having a separate variable for each structure.

For the second part of your review, I put the constants on top just to be consistent as all other structures have a similar variable. But I agree that its very intuitive to infer the size of each structure due to its name. If others feel the same, then we can go ahead and remove them.

harloff-g: I think I agree with Ben's point here. Given that every single static assert is doing the same thing & the size of the struct is in the name, I feel that there's enough indication of the purpose of the "magic" number that it would be fine not to have a variable.

basimsahaf: Is the removal of constants specific to the structures in this PR or for all other structures as well? (https://github.com/googleinterns/scsi2nvme/pull/20#discussion_r430522862)

eyhhuang: I also removed it from the other structures because I didn't see the value in extracting a constant since (I predict) we'll only be using that constant in the assert. Another benefit of _not_ having the constant is that you can see the number directly in the assert.

Ralith: I agree that the constants seem a bit superfluous in general. Given the existence of the static asserts, any downstream code that wants to rely on the size of the structs can just invoke `sizeof`, which is more readable anyway.

lib/scsi_defs.h
```diff
@@ -56,20 +57,36 @@ namespace scsi_defs {
   // Operation codes defined for SCSI commands supported by this project; this list may increase
   enum class OpCode : uint8_t {
     kTestUnitReady = 0x0,
+    kRead6 = 0x08,
+    kWrite6 = 0x0a,
     kRequestSense = 0x3,
     kInquiry = 0x12,
     kReserve6 = 0x16,
     kRelease6 = 0x17,
+    kRead10 = 0x28,
```
harloff-g, line 66: I think these are out of order?

lib/scsi_defs.h
```diff
@@ -56,20 +57,36 @@ namespace scsi_defs {
   // Operation codes defined for SCSI commands supported by this project; this list may increase
   enum class OpCode : uint8_t {
     kTestUnitReady = 0x0,
+    kRead6 = 0x08,
+    kWrite6 = 0x0a,
```
harloff-g, line 61: These come after kRequestSense

lib/scsi_defs.h
```diff
@@ -80,5 +97,282 @@ namespace scsi_defs {
...
+    uint8_t transfer_length: 8;
+    ControlByte control_byte;
+  } ABSL_ATTRIBUTE_PACKED;
+  static_assert(sizeof(Read6) == 6);
+  
+  // SCSI Reference Manual Table 97 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf  
+  struct Read10 {
+    OpCode op_code = OpCode::kRead10;
+    uint8_t rd_protect: 3;
+    bool dpo: 1;
```
harloff-g, line 114: Similar to https://github.com/googleinterns/scsi2nvme/pull/27#discussion_r430594682, should these have comments?

lib/scsi_defs.h
```diff
@@ -80,5 +97,282 @@ namespace scsi_defs {
     uint8_t vendorSpecific : 2;
   } ControlByte;
 
+  // SCSI Reference Manual Table 95 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct Read6 {
```
harloff-g, line 101: Nit: prefer to stay with a consistent style when defining command structs. In other PRs we have stuck to the standard of having all command structs suffixed with "Command" i.e. `Read6Command`

lib/scsi_defs.h
```diff
@@ -2,6 +2,7 @@
 #define LIB_SCSI_DEFS_H
 
 #include <cstdint>
+#include "absl/base/attributes.h"
```
harloff-g, line 5: nit: Newline between two types of includes

lib/scsi_defs.h
```diff
@@ -80,5 +97,282 @@ namespace scsi_defs {
...
+  } ABSL_ATTRIBUTE_PACKED;
+  static_assert(sizeof(Read10) == 10);
+
+  // SCSI Reference Manual Table 99 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf  
+  struct Read12 {
+    OpCode op_code = OpCode::kRead12;
+    uint8_t rd_protect: 3;
+    bool dpo: 1;
+    bool fua: 1;
+    bool rarc: 1;
```
harloff-g, line 134: nit: space between variable and colon

lib/scsi_defs.h
basimsahaf: Yup they were out of order. Fixed now

lib/scsi_defs.h
basimsahaf: Good catch. Will add comments!

lib/scsi_defs.h
```diff
@@ -80,5 +98,282 @@ namespace scsi_defs {
...
+    bool fua : 1; // Forced Unit access bit
+    bool rarc : 1; // Rebuild Assist Recovery bit
+    bool fua_nv : 1; // Forced Access Unit Non-Volatile 
+    bool obsolete : 1;
+    uint32_t logical_block_address : 32;
+    uint8_t reserved : 3;
+    uint8_t group_number : 5;
+    uint16_t transfer_length : 16;
+    ControlByte control_byte;
+
```
harloff-g, line 125: nit: remove extra newline

.gitignore
```diff
@@ -1,5 +1,5 @@
 #ignore all bazel build files
 /bazel-*
 
-#ignore local editor files
```
toadchild, line 4: Please move gitignore into its own change set, since it's not related to the rest of this,

basimsahaf:
> LGTM modulo [opcode discussion](https://github.com/googleinterns/scsi2nvme/pull/30#discussion_r432001349).

@Ralith 
Each struct in this case has a different opcode. Should we make a separate struct for other common fields in these structs?

.gitignore
basimsahaf: yup makes sense!

Ralith:
> Each struct in this case has a different opcode.

Sorry, I was unclear. I'm not concerned about the duplication of the fields here, but rather about having fields with hardcoded values on structures that I assume we're using for input. If we need to read the opcode to determine which structure to use, I'm not sure it's useful to have the opcode as a field in the structure, since it doesn't convey any information to any code that handles the structure type, and doesn't help us with decoding.

eyhhuang:
> Sorry, I was unclear. I'm not concerned about the duplication of the fields here, but rather about having fields with hardcoded values on structures that I assume we're using for input. If we need to read the opcode to determine which structure to use, I'm not sure it's useful to have the opcode as a field in the structure, since it doesn't convey any information to any code that handles the structure type, and doesn't help us with decoding.

So should we remove the hard-coded OpCode values?

Ralith:
My impression is that those fields won't be useful, though I could be missing something.

PR #30, opened 2020-05-26
Add Maintenance In structures

### LINKED ISSUE #15: Maintenance In command structures
Maintenance In command structures. Especially the part about support of op-codes.

harloff-g:
Closes #15 
Adds all POD structs relating to Maintenance in commands & data.


lib/scsi_defs.h
```diff
@@ -259,5 +259,131 @@ namespace scsi_defs {
   } ABSL_ATTRIBUTE_PACKED;
   static_assert(sizeof(ModeSense10Command) == 10);
 
+  
+  // SCSI Reference Manual Table 150 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct ReportOpCodesCommand {
+    OpCode op_code = OpCode::kMaintenanceIn;
+    uint8_t reserved_1 : 3;
+    uint8_t service_action : 5;
```
Ralith, line 267: In this and similar cases below, would it make sense to move the header into a separate struct, which could be used to guide decoding when the command isn't yet known?

harloff-g: This is definitely a good idea for these structures.

PR #31, opened 2020-05-26
Add Unmap Structures

### LINKED ISSUE #14: Unmap command structures

harloff-g:
Closes #14 
Adds three POD structs for the UNMAP command and response


lib/scsi_defs.h
```diff
@@ -259,5 +260,35 @@ namespace scsi_defs {
   } ABSL_ATTRIBUTE_PACKED;
   static_assert(sizeof(ModeSense10Command) == 10);
 
+  // SCSI Reference Manual Table 204 https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+  struct UnmapCommand {
+    OpCode op_code = OpCode::kUnmap;
```
Ralith, line 265: Would it be useful to decouple the opcode field from these structs? If we're using these mainly for decoding, it doesn't seem to helpful to include a field that we need to read *before* constructing the `UnmapCommand` from input. [Related discussion](https://github.com/googleinterns/scsi2nvme/pull/30#discussion_r432001349).

PR #37, opened 2020-05-29
Removing hardcoded opcodes values from master

basimsahaf:


harloff-g:
My impression was that Ben was suggesting that we remove the field entirely, not just the hard-coded value (https://github.com/googleinterns/scsi2nvme/pull/29#issuecomment-635552509).
What do you think?

basimsahaf:
> My impression was that Ben was suggesting that we remove the field entirely, not just the hard-coded value ([#29 (comment)](https://github.com/googleinterns/scsi2nvme/pull/29#issuecomment-635552509)).
> What do you think?

Yeah I think he meant that we should remove the fields altogether. It makes sense to me now that I think about it more. When we create a structure, we would seldom be using its opcode other than the time when we actually need to identify the type of struct we are supposed to create. In that scenario, we anyways won't be using this field at all and will simply refer to the opcode enum class. 

But I'm just wondering if there's a case where we need to do something like: `struct->op_code` in order to identify the type of struct. For example, this situation could occur when we are looping over an array of structs and only want to perform an operation on a specific type of struct.

eyhhuang:
Yeah similar to Basim, I kinda think it makes sense for the struct to hold explicit data about what type of command it is. Also, this is in line with the parsing approach where we essentially just bit cast the entire payload into our struct.

basimsahaf:
> Yeah similar to Basim, I kinda think it makes sense for the struct to hold explicit data about what type of command it is. Also, this is in line with the parsing approach where we essentially just bit cast the entire payload into our struct.

Yeah that's another reason but we can always leave out the opcode from the incoming payload and cast the rest of the data in case we decide on removing the fields altogether.

Ralith:
> But I'm just wondering if there's a case where we need to do something like: struct->op_code in order to identify the type of struct.

What is the type of `struct` in this example? That's the main issue here: the times when you need to know the opcode are, afaik, exactly the times when you do *not* already have it decoded into a struct. There's no way to construct an actual array directly containing different types.

> Yeah that's another reason but we can always leave out the opcode from the incoming payload and cast the rest of the data(since at that point we already know what struct we're trying to build) in case we decide on removing the fields altogether.

My expectation is that we'll have incoming data that contains an unknown operation; we'll then need to inspect the opcode to determine what struct to decode the remainder as.

basimsahaf:
That makes sense and I agree that it will be redundant. I'll go ahead remove the fields

basimsahaf:
Refer to #41

PR #39, opened 2020-05-29
Add style CI workflow. Fix all existing style errors.

### LINKED ISSUE #32: Check code style in CI
clang-format is a great tool for enforcing consistent style, both run manually to apply a style and in automated testing to ensure the style has been respected. Let's add a github actions job for the latter.

harloff-g:
Closes #32 

This PR has a lot of line changes. There were overarching style violations that changed almost all line indentations.


.clang-format
```diff
@@ -0,0 +1,173 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Google
```
Ralith, line 3: What's stopping us from using the Google defaults to make this file a lot more concise?

harloff-g: I've changed the workflow to generate the Google style format file dynamically (thus removing the need for this file at all). Do you think that's fine?

.github/workflows/main.yml
```diff
@@ -28,6 +28,10 @@ jobs:
       run: |
         sudo apt-get install clang-format
     
+    - name: Generate format guide
+      run: |
+        clang-format -style=google -dump-config > .clang-format
```
Ralith, line 33: Could we just have a committed .clang-format consisting of `BasedOnStyle:  Google` instead? That'd help people running the tool manually to format their own changes.

harloff-g: Thanks for pointing this out. I'm new to clang-format and didn't realize it was so easy

Issue #33, opened 2020-05-29
NVMe opcodes definitions


w-michal:
Self explanatory - we need NVMe OpCodes definitions.

Issue #34, opened 2020-05-29
Investigate importing NVMe definitions


w-michal:
There are already libraries (like SPDK) and headers that define all of that. Alternative approach here is to simply copy the structure definitions and modify them slightly to be more consistent with C++ style. Please research if open source google projects allow that and what other license types we can work with before proceeding.

harloff-g:
AFAIK we can work with the SPDK library. It is licences under BSD (https://github.com/spdk/spdk/blob/master/LICENSE), which is allowed for google open source projects with minimal restrictions other than including the notice/clause specified in the licence (go/thirdpartylicenses#notice).

PR #40, opened 2020-05-29
Nvme opcodes

eyhhuang:
Closes https://github.com/googleinterns/scsi2nvme/issues/33
Adds Admin opcodes, Command opcodes, and status codes


Ralith:
May want to consider whether it would save time/effort to import these from elsewhere, per #34.

eyhhuang:
> May want to consider whether it would save time/effort to import these from elsewhere, per #34.

That's a great consideration! The SPDK does in fact have all these definitions here https://github.com/spdk/spdk/blob/e10fc6ea00ba1281f6d48916fe6a33be5c10d80b/include/spdk/nvme_spec.h . However, their style is inconsistent with our C++ style guide so we would have to modify the code.

Ralith:
A bit of clang-format and search/replace can go a long way!

third_party/spdk_defs/nvme_defs.h
```diff
@@ -0,0 +1,210 @@
...
+  // Reserved 0x17
+  kHostidInconsistentFormat = 0x18,
+  kKeepAliveExpired = 0x19,
+  kKeepAliveInvalid = 0x1a,
+  kAbortedPreempt = 0x1b,
+  kSanitizeFailed = 0x1c,
+  kSanitizeInProgress = 0x1d,
+  kSglDataBlockGranularityInvalid = 0x1e,
+  kCommandInvalidInCmb = 0x1f,
+
```
harloff-g, line 59: nit: newlines between non-adjacent enum values is not consistent with master

third_party/spdk_defs/nvme_defs.h
```diff
@@ -0,0 +1,210 @@
+#ifndef LIB_NVME_DEFS_H
+#define LIB_NVME_DEFS_H
+
+#include <cstdint>
+
+#ifdef __cplusplus
+extern "C" {
```
harloff-g, line 7: If we want to build into a kernel module it is good to have this logic; however, master does not currently have it. Should we add it there?

third_party/spdk_defs/nvme_defs.h
eyhhuang: Good point!

third_party/spdk_defs/nvme_defs.h
eyhhuang: Yeah i think we should add it to scsi_defs as well

basimsahaf: would recommend having a separate PR for making it a Kernel module as it will be easier to roll back in the future in case we decide to not build this as a kernel module

eyhhuang: good idea! done

PR #51, opened 2020-06-03
NVMe Generic Queue Completion Entry

### LINKED ISSUE #50: Import Generic Completion Queue Entry
Completion Queue Entry (+ Status structure) must be formatted & imported from SPDK

harloff-g:
Closes #50 

Imports the generic queue completion entry struct and status field struct from SPDK.


third_party/spdk_defs/nvme_defs.h
```diff
@@ -192,7 +192,36 @@ enum class NvmOpcode : uint8_t {
...
+  uint32_t cdw0;  // command-specific
+  // dword 1
+  uint32_t rsvd1;
+  // dword 2
+  uint16_t sqhd;  // submission queue head pointer
+  uint16_t sqid;  // submission queue identifier
+  // dword 3
+  uint16_t cid;  // command identifier
+  spdk_nvme_status status;
+};
```
basimsahaf, line 221: Are we not doing struct packing for NVMe?

harloff-g: This particular structure doesn't need to be packed because it has no bit fields & has a size which is a multiple of it's largest type. I can add explicit bit fields and a packing tag if we feel it's more consistent, but when the structure has nothing fancy going on to begin with I think we might consider making an exception

basimsahaf: imo we should do it for consistency. Although others might feel differently because it doesn't make a difference on the compiler level.

Ralith: I'm not sure a well-defined memory layout is guaranteed unless packing is explicitly requested, so I'd favor doing at least that much.

third_party/spdk_defs/nvme_defs.h
```diff
@@ -0,0 +1,232 @@
...
+  kSctCommandSpecific = 0x1,
+  kSctMediaError = 0x2,
+  kSctPath = 0x3,
+  // Reserved 0x4-0x6
+  kSctVendorSpecific = 0x7,
+};
+
+// NVMe Base Specification Figure 15
+// https://nvmexpress.org/wp-content/uploads/NVM-Express-1_0e.pdf
+struct Status {
```
eyhhuang, line 199: is this status only for the completion queue entry? If so, what do you think of naming it something more specific so we don't accidentally get naming conflicts

third_party/spdk_defs/nvme_defs.h
```diff
@@ -0,0 +1,232 @@
...
+  kPdkNvmeSctGeneric = 0x0,
+  kSctCommandSpecific = 0x1,
+  kSctMediaError = 0x2,
+  kSctPath = 0x3,
+  // Reserved 0x4-0x6
+  kSctVendorSpecific = 0x7,
+};
+
+// NVMe Base Specification Figure 15
+// https://nvmexpress.org/wp-content/uploads/NVM-Express-1_0e.pdf
```
eyhhuang, line 198: Is the reason that you used Revision 1.0e because that's what spdk uses?

harloff-g: Good question! I actually didn't consider what version I was linking here. I've noticed that all other links go to the 1_4 2019 spec of NVMe, but there's an updated 2020 spec 1_4a. I couldn't find a direct reference to what SPDK uses, but it may be the case that they update the library to always support features in the newest version. Is there a reason we chose the 2019 spec? https://nvmexpress.org/wp-content/uploads/NVM-Express-1_4a-2020.03.09-Ratified.pdf

third_party/spdk_defs/nvme_defs.h
harloff-g: Yeah, it's just a completion entry thing. What do you think about naming it "CplStatus"

eyhhuang: good with me!

third_party/spdk_defs/nvme_defs.h
eyhhuang: I believe the 2019 spec was listed in our project's google doc, so it might be worth asking the FTEs for this. But since the 2020 is just 1_4a, I don't imagine it'll be too different from the 2019 1_4 one

eyhhuang: I've been using the 2019 version to double check the SPDK implementation and then just leaving a comment where some fields were not included in the SPDK code

harloff-g: Alright, I'll switch to using 1_4 for all my docs

Issue #52, opened 2020-06-04
Import Identify controller


eyhhuang:
Some of subfields in these structures could be its own struct, like “SGL support Field”, “Async Event Support”, “Volatile write cache”, “Support for flush all”, “Optional NVM command support”, “Log Page attributes”, “Controller attributes”, “Power state descriptors”, Enum that specifies type of identify (namespace/controller/active Napesaces ID list etc).

PR #55, opened 2020-06-05
Identify controller

eyhhuang:
Closes https://github.com/googleinterns/scsi2nvme/issues/52


third_party/spdk_defs/nvme_defs.h
```diff
@@ -239,6 +239,295 @@ struct SglDescriptor {
...
+  uint8_t reserved3 : 3;
+  uint8_t rrl : 5;  // bits 108:104: relative read latency
+  uint8_t reserved4 : 3;
+  uint8_t rwt : 5;  // bits 116:112: relative write throughput
+  uint8_t reserved5 : 3;
+  uint8_t rwl : 5;  // bits 124:120: relative write latency
+  uint8_t reserved6 : 3;  // includes fields added in NVMe Revision 1.4
+  uint8_t reserved7[16];  // includes fields added in NVMe Revision 1.4
+} ABSL_ATTRIBUTE_PACKED;
+static_assert(sizeof(struct PowerState) == 32);
```
harloff-g, line 275: Don't need to append struct to the start of the type in c++

third_party/spdk_defs/nvme_defs.h
```diff
@@ -239,6 +239,295 @@ struct SglDescriptor {
...
+    uint8_t ctrlr_model : 1;
+    uint8_t reserved : 7;
+  } ctrattr;  // Controller attributes
+
+  uint8_t msdbd;  // Maximum SGL block descriptors (0 = no limit)
+  uint8_t reserved[244];
+} ABSL_ATTRIBUTE_PACKED;
+static_assert(sizeof(IdentifyControllerNvmfSpecific) == 256);
+
+constexpr uint8_t kCtrlrSnLen = 20;  // controller serial number length
```
harloff-g, line 296: I'd prefer defining constants at the start of the namespace for organizational purposes but I'm interested in what others think is the best approach.

third_party/spdk_defs/nvme_defs.h
```diff
@@ -239,6 +239,295 @@ struct SglDescriptor {
...
+      uint8_t reserved : 7;
+    } bits;
+  } dsto;  // device self-test options
+
+  // Firmware update granularity
+  //  4KB units
+  // 0x00 = no information provided
+  // 0xFF = no restriction
+  uint8_t fwug : 8;
+
```
harloff-g, line 430: nit: remove space between these two basic fields

third_party/spdk_defs/nvme_defs.h
```diff
@@ -239,6 +239,295 @@ struct SglDescriptor {
...
+constexpr uint8_t kCtrlrFrLen = 8;   // controller firmware revision length
+constexpr uint16_t kNqnFieldSize = 256;  // NVMe qualified name field size
+
+// NVMe base specification Figure 247
+// https://nvmexpress.org/wp-content/uploads/nvm-express-1_4-2019.06.10-ratified.pdf
+struct IdentifyControllerData {
+  // bytes 0-255: controller capabilities and features
+  uint16_t vid : 16;        // pci vendor id
+  uint16_t ssvid : 16;      // pci subsystem vendor id
+  int8_t sn[kCtrlrSnLen];   // serial number
```
harloff-g, line 307: I'm not sure the usefulness of defining constants and then just using them once to define array length. Do you know if the constants will have further use later?

third_party/spdk_defs/nvme_defs.h
```diff
@@ -239,6 +239,295 @@ struct SglDescriptor {
...
+union VsRegister {
+  uint32_t raw;
+  struct {
+    uint32_t ter : 8;   // indicates the tertiary version
+    uint32_t mnr : 8;   // indicates the minor version
+    uint32_t mjr : 16;  // indicates the major version
+                        // reserved for 1.1, 1.2, 1.2.1 Compliant Controllers
+  } bits;
+} ABSL_ATTRIBUTE_PACKED;
+static_assert(sizeof(union VsRegister) == 4);
```
harloff-g, line 253: AFAIK you don't need to append union when referencing a union type in c++

third_party/spdk_defs/nvme_defs.h
eyhhuang: good catch

third_party/spdk_defs/nvme_defs.h
eyhhuang: i looked through spdk and they do use the constants again in the `spdk_nvmf_subsystem' struct. However, I'm not sure whether we are likely to re-use the constants. One idea is that since these constants give information about the constraints/properties of some fields, they might be useful if we have to do input validation.

third_party/spdk_defs/nvme_defs.h
eyhhuang: that's a good point. My reasoning was that I wanted the constant to be close to the struct where it is used. However, if the constant will be used in multiple places then putting the constant near one usage of it no longer makes sense, so it'd makes sense to put it at the start of the namespace. And if the constant is only used in one place, then perhaps it doesn't make sense to have a constant at all.

third_party/spdk_defs/nvme_defs.h
eyhhuang: going to remove constants because we can add them back later if they're used in more than one place

third_party/spdk_defs/nvme_defs.h
```diff
@@ -270,6 +270,289 @@ struct SglDescriptor {
...
+    // Supports sending Namespace Attribute Notices
+    uint32_t ns_attribute_notices : 1;
+    // Supports sending Firmware Activation Notices
+    uint32_t fw_activation_notices : 1;
+    uint32_t reserved2 : 22;
+  } oaes;  // optional asynchronous events support
+
+  struct {
+    uint32_t host_id_exhid_supported : 1;  // Supports 128-bit host identifier
+    uint32_t
```
harloff-g, line 363: nit: variable & type on single line

PR #56, opened 2020-06-05
Add third_party directory to style check CI

harloff-g:
Title


harloff-g:
> I'd lean towards not doing this, because `third_party` dirs typically contain complete external projects, for which reformatting would be an unreasonable maintenance burden. `spdk_defs` is a bit of a special case because we've pulled out only the data we specifically want, and there's little need to sync it with upstream spdk in the future since it's all defined by the NVMe spec regardless. Maybe just target the spdk_defs directory directly?

Good point, I can see how enforcing style on entire libraries is an unreasonable thing to do. Enforcing style on code that we are intentionally styling seems relevant though, so I've updated the PR to only check style on spdk_defs.

Issue #60, opened 2020-06-12
Translate Status Codes


eyhhuang:
Translate NVMe status codes to SCSI status code, sense key, and additional sense code

PR #62, opened 2020-06-15
Translate status code

eyhhuang:
Closes https://github.com/googleinterns/scsi2nvme/issues/60


lib/translator.h
```diff
@@ -0,0 +1,26 @@
+#ifndef LIB_TRANSLATOR_H
+#define LIB_TRANSLATOR_H
+
+#include "scsi_defs.h"
+#include "../third_party/spdk_defs/nvme_defs.h"
+
+namespace translator {
+struct TranslatedScsiStatus {
```
harloff-g, line 8: Thoughts on calling this just "ScsiStatus" rather than "TranslatedScsiStatus"? It seems that the general use case for this structure is more generic than only holding translated data & name is more succinct

eyhhuang: I'd be happy to go with `ScsiStatus`. My only concern is that that might be confused with `scsi_defs::Status` https://github.com/googleinterns/scsi2nvme/blob/master/lib/scsi_defs.h#L29

harloff-g:
Just some general thoughts about how this is set up, but what do you think about having a single directory for all translation code called "translate". In this directory, there will be a .h and .cc file for every command (or discrete entity) that can be translated containing the translation code. All of these things can be in one namespace named "translator". This way, we group all of the translation code together while also splitting it up & we don't have a ton of namespaces to remember

eyhhuang:
> Just some general thoughts about how this is set up, but what do you think about having a single directory for all translation code called "translate". In this directory, there will be a .h and .cc file for every command (or discrete entity) that can be translated containing the translation code. All of these things can be in one namespace named "translator". This way, we group all of the translation code together while also splitting it up & we don't have a ton of namespaces to remember

I like the idea of having a single directory called "translate". WRT the namespace, would they all be under the translate namespace and we have differently named translation functions, e.g. `translate::translateStatusCode` or would they have different namespaces and all have a translate function, e.g. `status_translator::translate`

lib/translate/BUILD
```diff
@@ -0,0 +1,11 @@
+cc_library(
+  name = "status_translator_lib",
```
eyhhuang, line 2: Not sure whether the best option is to have separate libraries for each translator or to just have one translator library.

harloff-g: I vote for having them all in one library called "translator_lib"

lib/translate/status_translator.cc
```diff
@@ -0,0 +1,240 @@
...
+      sense_key = scsi_defs::SenseKey::kIllegalRequest;
+      sense_code = std::optional<scsi_defs::AdditionalSenseCode>{
+          scsi_defs::AdditionalSenseCode::kInvalidFieldInCdb};
+      break;
+    default:
+      printf("No SCSI translation for NVMe status with code %#x\n",
+             (uint8_t)sc);
+      return nullptr;
+  }
+  return new TranslatedScsiStatus(status, sense_key, sense_code);
```
harloff-g, line 177: Given recent memory discussions we may want to offload this allocation to outside of this file and have a pointer to the allocated struct as a parameter to the translation method

eyhhuang: good point!

lib/translate/common.h
```diff
@@ -0,0 +1,7 @@
+#ifndef TRANSLATION_COMMON_LIB
+#define TRANSLATION_COMMON_LIB
+
+#define TRANSLATION_SUCCESS 1
+#define TRANSLATION_FAIL 0
```
Ralith, line 5: An `enum class` would be more idiomatic here.

test/status_translator_test.cc
```diff
@@ -0,0 +1,100 @@
...
+
+// Tests
+
+namespace {
+
+TEST(TranslateGenericCommandStatus, ShouldReturnCorrectStatus) {
+  uint8_t sct = 0x0;  // kGeneric
+  uint8_t sc = 0x0;   // kSuccess;
+  translator::TranslatedScsiStatus *translated =
+      new translator::TranslatedScsiStatus();
```
Ralith, line 29: It's usually a bad idea in modern C++ to ever call `new` explicitly, as it makes it easy to inadvertently create a memory leak, as is the case here. Instead you can use something like `unique_ptr` which takes care of cleanup automatically.

However, in this particular case we don't need a heap allocation at all; you can define a `TranslatedScsiStatus` on the stack and pass in it address. This is nearly always preferred when feasible, since it's faster and less error-prone.

eyhhuang: good general tip about the `unique_ptr`! and you're definitely right that here a stack allocated struct is sufficient

lib/translate/status_translator.h
```diff
@@ -0,0 +1,46 @@
...
+};
+
+int translateStatusCode(const uint8_t sct, const uint8_t sc,
+                        TranslatedScsiStatus *translated);
+int translateStatusCode(const nvme_defs::GenericCommandStatusCode sc,
+                        TranslatedScsiStatus *translated);
+int translateStatusCode(const nvme_defs::CommandSpecificStatusCode sc,
+                        TranslatedScsiStatus *translated);
+int translateStatusCode(const nvme_defs::MediaErrorStatusCode sc,
+                        TranslatedScsiStatus *translated);
```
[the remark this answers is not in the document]
JustinAWei: @Ralith is `std::optional` easily available in kernel modules?

lib/translate/status_translator.cc
```diff
@@ -0,0 +1,234 @@
...
+    case nvme_defs::StatusCodeType::kMediaError:
+      return translateStatusCode(
+          static_cast<nvme_defs::MediaErrorStatusCode>(sc), translated);
+    case nvme_defs::StatusCodeType::kPath:
+    case nvme_defs::StatusCodeType::kVendorSpecific:
+    default:
+      printf(
+          "No SCSI translation for NVMe status code type %#x"
+          "and status code %#x\n",
+          sct, sc);
```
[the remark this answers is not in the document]
eyhhuang: Do you mean something like storing the error message in a global string and having that be accessible everywhere? I'm not too sure how the debug callback is supposed to work.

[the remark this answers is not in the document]
Ralith: The debug callback would be a function pointer which the translation library invokes with messages as needed. For example:

```c++
static void (*debug_callback)(char *);
```

We'd want to export a setter function to fill it in with a user-supplied value, and check whether it's nonnull before invoking it.

lib/translate/status_translator.h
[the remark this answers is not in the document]
Ralith: That's a good question; if it's not, the definition of `TranslatedScsiStatus` will need to be adjusted as well. I suspect we might get away with it since, being a template, it's unlikely to pull in any standard library runtime symbols, but some research and/or experimentation will be needed to confirm.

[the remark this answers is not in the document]
harloff-g: #71 for follow-ups on this discussion

lib/translate/status_translator.h
```diff
@@ -0,0 +1,46 @@
...
+  util::Optional<scsi_defs::AdditionalSenseCode> sense_code;
+
+  TranslatedScsiStatus() {}
+  TranslatedScsiStatus(
+      scsi_defs::Status status, scsi_defs::SenseKey sense_key,
+      util::Optional<scsi_defs::AdditionalSenseCode> sense_code)
+      : status{status}, sense_key{sense_key}, sense_code{sense_code} {}
+};
+
+util::Optional<TranslatedScsiStatus> translateStatusCode(const uint8_t sct,
```
harloff-g, line 36: is marking passed-by-value integers as constant good practice?

eyhhuang: Since the const doesn't make a difference for passing by value, probably not? This was done out of habit for marking as const when passing by references. We can remove this

lib/translate/common.h
```diff
@@ -0,0 +1,24 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIB_TRANSLATOR_COMMON_H
+#define LIB_TRANSLATOR_COMMON_H
+
+static void (*debug_callback)(const char*);
```
Ralith, line 18: This will define a separate, unique global variable in every file that includes it. To have a single global, place the definition inside a `.cc` file, without the `static` qualifier that makes it internal to the compilation unit. The header file should only contain a *declaration*, which indicates to including code that the global exists in another compilation unit. In this case a declaration would use the same syntax as the definition, but with the `extern` qualifier.

However, in this particular case we should go a step further. Exposing the global allows any including code to write to it, but for readability's sake we should ensure that the only way to write to the global is with the provided setter function. To achieve that, omit the global's declaration from the header entirely, and define a helper function named something like `DebugLog` for use within this library's code which just forwards its arguments to the function pointer, if set. That will also save downstream code from having to check whether the pointer's set everywhere.

lib/translate/status_translator.h
```diff
@@ -0,0 +1,46 @@
...
+};
+
+util::Optional<TranslatedScsiStatus> translateStatusCode(uint8_t sct,
+                                                         uint8_t sc);
+util::Optional<TranslatedScsiStatus> translateStatusCode(
+    nvme_defs::GenericCommandStatusCode sc);
+util::Optional<TranslatedScsiStatus> translateStatusCode(
+    nvme_defs::CommandSpecificStatusCode sc);
+util::Optional<TranslatedScsiStatus> translateStatusCode(
+    nvme_defs::MediaErrorStatusCode sc);
```
Ralith, line 43: [Functions should be named in PascalCase](https://google.github.io/styleguide/cppguide.html#Function_Names).

lib/translate/status_translator.h
Ralith: Yeah, `const` here doesn't do anything at all and hence is best omitted for concision.

lib/translate/status_translator.h
eyhhuang: I'll try to be more careful with styling in the future!

lib/translate/status_translator.cc
```diff
@@ -0,0 +1,257 @@
...
+      return translateStatusCode(
+          static_cast<nvme_defs::CommandSpecificStatusCode>(sc));
+    case nvme_defs::StatusCodeType::kMediaError:
+      return translateStatusCode(
+          static_cast<nvme_defs::MediaErrorStatusCode>(sc));
+    case nvme_defs::StatusCodeType::kPath:
+    case nvme_defs::StatusCodeType::kVendorSpecific:
+    default:
+      char debug_buffer[100];
+      sprintf(debug_buffer,
```
JustinAWei, line 40: can we make this into a general fn?

Ralith: This is an excellent idea. The `DebugLog` function I proposed [above](https://github.com/googleinterns/scsi2nvme/pull/62#discussion_r445142974) could be made variadic and call `snprintf` internally. Something more iostreams-style could also be done, but that'd probably be a lot more effort.

We should also double-check that `snprintf` is available in the kernel, and find an alternative approach if not.

toadchild: Yes, linux/include/linux/kernel.h includes snprintf.

lib/translate/common.h
eyhhuang: Thank you for the thorough explanation! I think I understand much better how the debug callback should work now :)

lib/translate/BUILD
```diff
@@ -0,0 +1,11 @@
+cc_library(
+  name = "translator_lib",
+  srcs = ["common.cc", "status_translator.cc"],
```
oicirtap, line 3: A better approach for the BUILD file would be to create a cc_library target for each translator component; in this case, one `status_translator` target, and one `common` target, as well as a target for each other command (inquiry, read, etc). Then we would have a translator target which brings everything together.

Something like this:

```
cc_library(
  name = "common",
  hdrs = ["common.h"],
  srcs = ["common.cc"],
  deps = [
      "@com_google_absl//absl/base",
  ],
)

cc_library(
  name = "status_translator",
  hdrs = ["status_translator.h"],
  srcs = ["status_translator.cc"],
  deps = [
      ":common"
      "//third_party/spdk_defs:nvme_defs_lib",
      "//lib:scsi_defs_lib",
      "@com_google_absl//absl/base",
  ],
)
```

This approach allows modular development, helps keep build rules small and clean, and is in general easier to follow.

@eyhhuang @harloff-g @JustinAWei @basimsahaf I would encourage you to talk about this and discuss the specifics of how you want to structure the BUILD file.

test/BUILD
```diff
@@ -1,10 +1,11 @@
 cc_test(
```
oicirtap, line 1: In the same spirit as my suggestion above, I encourage you to have a test target for each individual component (hello_world, scsi_defs, nvme_defs, status_translator, etc). This will allow you to run only the tests relevant to the piece you are currently working on. You can always run all of your tests with `bazel test :all`

For status_translator, this could look like this:

```
cc_library(
  name = "status_translator_test",
  srcs = ["status_translator_test.cc"],
  deps = [
      "//lib/translate:status_translator"
      "//third_party/spdk_defs:nvme_defs_lib",
      "//lib:scsi_defs_lib",
      "@googletest//:gtest_main",
      "@com_google_absl//absl/base",
  ],
)
```

test/optional_test.cc
```diff
@@ -0,0 +1,49 @@
+// Copyright 2020 Google LLC
```
oicirtap, line 1: Similar to common.h / common.c, This could be in its own PR it you wished. Since the optional tests are not blocking other PRs I'll leave it up to you.

test/BUILD
```diff
@@ -1,10 +1,11 @@
...
     name = "tests",
-    srcs = glob(["**/*.cc"]),
```
oicirtap, line 3: I would personally avoid using such generic regex expressions for build targets. This can lead to future files unintentionally being added to this target (Whenever someone adds a file in this directory, they would need to check if it doesn't affect this target, even if they didn't intend to change any existing modules).

With my above suggestion of having many build targets for tests, this should not be necessary.

lib/translate/common.h
```diff
@@ -0,0 +1,36 @@
+// Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
eyhhuang: I definitely agree! @harloff-g is currently working to separate out a common.h into its own PR and we can rebase off of that. This PR definitely ballooned to include a lot of changes so I'll make sure to separate out components in the future!

lib/translate/common.h
```diff
@@ -0,0 +1,36 @@
...
+
+void SetDebugCallback(void (*callback)(const char*));
+void DebugCallback(const char* buffer);
+
+template <typename... Args>
+void DebugLog(const char* format, Args... args) {
+  char buffer[100];
+  snprintf(buffer, 100, format, args...);
+  DebugCallback(buffer);
+}
```
Ralith, line 32: Rather than using a variadic template, `DebugCallback` itself could be a C-style variadic function which forwards its arguments to `vsnprintf`. I'd also lean towards a bit more buffer space--say 1024 bytes.

lib/translate/BUILD
oicirtap: I added a more extensive comment in a different thread, but I think the cleaner approach is to have each translator in its own cc_library target, and then have one big cc_library target that includes all of the individual targets.

lib/translate/BUILD
eyhhuang: Thanks for all your BUILD structure suggestions! Definitely a good idea to keep things modularized and clean, especially as the codebase expands. We'll split up the tests and libraries into their own targets

lib/translator.h
eyhhuang: Ultimately decided to name this ScsiStatus. The different namespaces should be a sufficient differentiator.

lib/translate/BUILD
eyhhuang: https://github.com/googleinterns/scsi2nvme/pull/83

lib/BUILD
```diff
@@ -1,7 +1,7 @@
 cc_library(
```
eyhhuang, line 1: BUILD changes to master have been covered in this PR https://github.com/googleinterns/scsi2nvme/pull/83

lib/translator/status.cc
```diff
@@ -0,0 +1,252 @@
...
+    case nvme_defs::GenericCommandStatusCode::kNamespaceNotReady:
+      result.status = scsi_defs::Status::kCheckCondition;
+      result.sense_key = scsi_defs::SenseKey::kNotReady;
+      result.asc = scsi_defs::AdditionalSenseCode::
+          kLogicalUnitNotReadyCauseNotReportable;
+      result.ascq = scsi_defs::AdditionalSenseCodeQualifier::
+          kLogicalUnitNotReadyCauseNotReportable;
+      break;
+    default:
+      DebugLog("No SCSI translation for NVMe status with code %#x\n",
```
Ralith, line 102: Let's omit trailing newlines from `DebugLog` invocations; downstream code can add them when appropriate, which isn't always.
```suggestion
      DebugLog("No SCSI translation for NVMe status with code %#x",
```

lib/translator/status.cc
```diff
@@ -0,0 +1,252 @@
...
+          scsi_defs::AdditionalSenseCodeQualifier::kFormatCommandFailed;
+      break;
+    case nvme_defs::CommandSpecificStatusCode::kConflictingAttributes:
+      result.status = scsi_defs::Status::kCheckCondition;
+      result.sense_key = scsi_defs::SenseKey::kIllegalRequest;
+      result.asc = scsi_defs::AdditionalSenseCode::kInvalidFieldInCdb;
+      result.ascq = scsi_defs::AdditionalSenseCodeQualifier::kInvalidFieldInCdb;
+      break;
+    default:
+      DebugLog("No SCSI translation for NVMe status with code %#x\n",
```
Ralith, line 152:
```suggestion
      DebugLog("No SCSI translation for NVMe status with code %#x",
```

lib/translator/status.cc
```diff
@@ -0,0 +1,252 @@
...
+      result.sense_key = scsi_defs::SenseKey::kNotReady;
+      result.asc = scsi_defs::AdditionalSenseCode::
+          kLogicalUnitNotReadyCauseNotReportable;
+      result.ascq = scsi_defs::AdditionalSenseCodeQualifier::
+          kLogicalUnitNotReadyCauseNotReportable;
+      break;
+    default:
+      DebugLog("No SCSI translation for NVMe status with code %#x\n",
+               static_cast<uint8_t>(sc));
+      return StatusCode::kNoTranslation;
```
Ralith, line 104: What is downstream code expected to do when this case (or a similar one below) arises, since no status/sense_key is supplied? Should we set some catch-all values so execution can proceed?

lib/translator/status.h
```diff
@@ -0,0 +1,34 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIB_TRANSLATOR_STATUS_H
+#define LIB_TRANSLATOR_STATUS_H
+
+#include "common.h"
```
JustinAWei, line 18: nit: order goes like this.
Other libraries' .h files.
Your project's .h files.

should third_party be "other libraries' h files" and go first?

test/translator/status_test.cc
```diff
@@ -0,0 +1,97 @@
...
+  translator::StatusCode status = translator::Status(sct, sc, result);
+
+  EXPECT_EQ(status, translator::StatusCode::kSuccess);
+  EXPECT_EQ(result.status, scsi_defs::Status::kGood);
+  EXPECT_EQ(result.sense_key, scsi_defs::SenseKey::kNoSense);
+  EXPECT_EQ(result.asc, scsi_defs::AdditionalSenseCode::kNoAdditionalSenseInfo);
+  EXPECT_EQ(result.ascq,
+            scsi_defs::AdditionalSenseCodeQualifier::kNoAdditionalSenseInfo);
+}
+TEST(TranslateUnsupportedGenericCommandStatus, ShouldReturnNullptr) {
```
JustinAWei, line 36: nit: newline

lib/translator/status.cc
```diff
@@ -0,0 +1,252 @@
...
+StatusCode Status(nvme_defs::CommandSpecificStatusCode sc, ScsiStatus& result) {
+  result.asc = scsi_defs::AdditionalSenseCode::kNoAdditionalSenseInfo;
+  result.ascq = scsi_defs::AdditionalSenseCodeQualifier::kNoAdditionalSenseInfo;
+
+  switch (sc) {
+    case nvme_defs::CommandSpecificStatusCode::kCompletionQueueInvalid:
+      result.status = scsi_defs::Status::kCheckCondition;
+      result.sense_key = scsi_defs::SenseKey::kIllegalRequest;
+      break;
+    case nvme_defs::CommandSpecificStatusCode::
```
toadchild, line 120: These line breaks look awkward to me; would it make sense to put the entire identifier on one line and the "unspecified" comments above/below?

eyhhuang: I agree the one line approach looks cleaner. However, the git-clang-format automatically reformats it to be on two lines. We could potentially increase the maximum line length for clang-format (if possible) to get around this

lib/translator/status.cc
```diff
@@ -0,0 +1,252 @@
...
+
+#include "status.h"
+
+namespace translator {
+
+namespace {  // anonymous namespace for helper functions
+
+// Section 7.1
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+StatusCode Status(nvme_defs::GenericCommandStatusCode sc, ScsiStatus& result) {
```
toadchild, line 23: Should the name of this family of functions indicate that it's translating from NVMe to SCSI?

lib/translator/status.cc
eyhhuang: That's a good concern! For NVMe, there is only one status code for success, so we know that all others are some type of failure. Therefore, we can technically set the `SenseKey`, `AdditionalSenseCode`, etc. values to be some generic failure. However, generic failure codes don't really exist. 

As it currently is,  when there is no NVMe to SCSI translation, the `translator::ScsiStatus` object will remain default constructed, in which case both `SenseKey` and `AdditionalSenseCode` will be `kNoSense`. In this case, downstream code, after encountering a `translator::StatusCode::kNoTranslation`, can still proceed with the execution and send the various `kNoSense` codes back to the SCSI driver

lib/translator/status.cc
eyhhuang: That's a good point. As it is, it's unclear that the translation is from NVMe to SCSI, so i'll rethink the naming and consult the others about this

lib/translator/status.h
eyhhuang: I'm not too sure what counts as "other libraries" bc the scsi_defs is technically in its own library. However, when I try to rearrange the imports, clang-format changes them to be in alphabetical order

lib/translator/status.cc
toadchild: Ah, I meant possibly moving the // unspecified comments to not be every line; if you've got a big fall-through block, one comment is probably enough.

eyhhuang: Ah gotcha! Yep, I think doing so would still ensure it's clear that those statuses are unspecified and will look cleaner

lib/translator/status.h
Ralith: Separate different categories of headers with an empty line. This is more readable for large groups of headers and will make clang-format behave.

eyhhuang: Hmm, oddly enough even when I put `common.h` on a newline, it puts it back to the same spot.
image

lib/translator/status.cc
eyhhuang: Renamed to `StatusToScsi()`

lib/translator/status.h
Ralith: Ah, it looks like we have a clang-format configuration issue. The Google style default assumes `IncludeCategories` has been configured suitably to distinguish the groups automatically, but we haven't done that. We should extend out `.clang-format` with either `IncludeBlocks: Preserve` or by defining an include category for `third_party`.

Ralith: Perfectly fine to consider that out of scope for this PR, though, IMO.

eyhhuang: https://github.com/googleinterns/scsi2nvme/issues/94 created an issue to address this

test/translator/status_test.cc
```diff
@@ -0,0 +1,98 @@
...
+#include "gtest/gtest.h"
+
+// Tests
+
+namespace {
+
+TEST(TranslateGenericCommandStatus, ShouldReturnCorrectStatus) {
+  uint8_t sct = 0x0;  // kGeneric
+  uint8_t sc = 0x0;   // kSuccess;
+  translator::ScsiStatus result;
```
harloff-g, line 26: Since we rely on zero-init construction at line 103 in status.cc, this might need to be translator::ScsiStatus result = {}.

eyhhuang: I'm changing status.cc to not require zero initialization. Instead, status.cc will explicitly set `ScsiStatus` to the "default" type of kCheckCondition and kNoSense when there's no translation available

lib/translator/status.cc
```diff
@@ -0,0 +1,243 @@
...
+
+#include "common.h"
+
+namespace translator {
+
+namespace {  // anonymous namespace for helper functions
+
+// Section 7.1
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+void StatusToScsi(nvme_defs::GenericCommandStatusCode sc, ScsiStatus& result) {
```
w-michal, line 25: nit: prefer return values over mutable func parameters.

lib/translator/status.cc
```diff
@@ -0,0 +1,243 @@
...
+      result.status = scsi_defs::Status::kCheckCondition;
+      result.sense_key = scsi_defs::SenseKey::kNotReady;
+      result.asc = scsi_defs::AdditionalSenseCode::
+          kLogicalUnitNotReadyCauseNotReportable;
+      result.ascq = scsi_defs::AdditionalSenseCodeQualifier::
+          kLogicalUnitNotReadyCauseNotReportable;
+      break;
+    default:
+      DebugLog("No SCSI translation for NVMe status with code %#x",
+               static_cast<uint8_t>(sc));
```
Ralith, line 102: The caller has no way to detect this case (and all the similar fallbacks below), so we need to either fill in the output with *something* or add a return value indicating whether the result was written which the next layer up will have to do something with.

From a high level viewpoint, if we're translating NVMe to SCSI, we're going to have to come up with *some* response no matter what strange condition arises, otherwise the SCSI user won't be able to make progress.

lib/translator/status.cc
eyhhuang: these are just helper functions and not part of the interface. Nonetheless, would you still prefer I make the change?

Ralith: It's the preferred style whether or not it's public-facing, so long as it's practical to accomplish. So long as we're not returning a status code or similar here it'll make the code more readable.

eyhhuang: Okay, good to know! I'll make the change

lib/translator/status.cc
```diff
@@ -20,9 +20,20 @@ namespace translator {
...
 namespace {  // anonymous namespace for helper functions
 
+ScsiStatus BuildDefaultScsiStatus() {
+  return ScsiStatus{
+      .status = scsi_defs::Status::kCheckCondition,
+      .sense_key = scsi_defs::SenseKey::kNoSense,
+      .asc = scsi_defs::AdditionalSenseCode::kNoAdditionalSenseInfo,
+      .ascq = scsi_defs::AdditionalSenseCodeQualifier::kNoAdditionalSenseInfo,
+  };
+}
```
Ralith, line 30: Could this be a constant instead of a function?

lib/translator/status.h
```diff
@@ -0,0 +1,42 @@
...
+
+  bool operator==(const ScsiStatus& other) const;
+};
+
+constexpr ScsiStatus kDefaultScsiStatus = ScsiStatus{
+    .status = scsi_defs::Status::kCheckCondition,
+    .sense_key = scsi_defs::SenseKey::kNoSense,
+    .asc = scsi_defs::AdditionalSenseCode::kNoAdditionalSenseInfo,
+    .ascq = scsi_defs::AdditionalSenseCodeQualifier::kNoAdditionalSenseInfo,
+};
```
Ralith, line 37: `constexpr`s appearing in headers should be marked `inline` or else they're a severe correctness hazard: if they're not `inline` then any use of them inside an `inline` function that gets compiled into multiple compilation unit is undefined behavior!

Do we actually need this to be in the header (i.e. visible to other compilation units) at all, though?

eyhhuang: I was thinking they might be useful if any function wants to check whether the `ScsiStatus` is just the default value. Although, we could keep this just in the .cc for now and expand its visibility when the need arises.

lib/translator/status.cc
```diff
@@ -0,0 +1,248 @@
...
+constexpr ScsiStatus kDefaultScsiStatus{
+    .status = scsi_defs::Status::kCheckCondition,
+    .sense_key = scsi_defs::SenseKey::kNoSense,
+    .asc = scsi_defs::AdditionalSenseCode::kNoAdditionalSenseInfo,
+    .ascq = scsi_defs::AdditionalSenseCodeQualifier::kNoAdditionalSenseInfo,
+};
+
+// Section 7.1
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+ScsiStatus StatusToScsi(nvme_defs::GenericCommandStatusCode sc) {
```
oicirtap, line 32: It's good practice to avoid abbreviations for variable names. "Do not abbreviate by deleting letters within a word."
https://google.github.io/styleguide/cppguide.html#Naming

Let's change `sc` to `status_code`.

lib/translator/status.cc
```diff
@@ -0,0 +1,248 @@
...
+    case nvme_defs::CommandSpecificStatusCode::kAbortCommandLimitExceeded:
+    case nvme_defs::CommandSpecificStatusCode::kAsyncEventRequestLimitExceeded:
+    case nvme_defs::CommandSpecificStatusCode::kInvalidFirmwareSlot:
+    case nvme_defs::CommandSpecificStatusCode::kInvalidFirmwareImage:
+    case nvme_defs::CommandSpecificStatusCode::kInvalidInterruptVector:
+    case nvme_defs::CommandSpecificStatusCode::kInvalidLogPage:
+      // Translation unspecified
+      result.status = scsi_defs::Status::kCheckCondition;
+      result.sense_key = scsi_defs::SenseKey::kIllegalRequest;
+      break;
```
oicirtap, line 137: Here, the code for `kCompletionQueueInvalid` is the same as for the next clumped cases. This seems deliberate (I'm guessing to differentiate the explicit conversion from undefined ones?). I would extend your `// Translation unspecified` comment (or add a new comment above) to call out why `kCompletionQueueInvalid` gets its own section, even when the actual logic is equivalent.

lib/translator/status.cc
```diff
@@ -0,0 +1,248 @@
...
+    default:
+      DebugLog("No SCSI translation for NVMe status with code %#x",
+               static_cast<uint8_t>(sc));
+  }
+  return result;
+}
+
+// Section 7.2
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+ScsiStatus StatusToScsi(nvme_defs::CommandSpecificStatusCode sc) {
```
oicirtap, line 118: ditto `sc` -> `status_code`

lib/translator/status.cc
```diff
@@ -0,0 +1,248 @@
...
+    default:
+      DebugLog("No SCSI translation for NVMe status with code %#x",
+               static_cast<uint8_t>(sc));
+  }
+  return result;
+}
+
+// Section 7.3
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+ScsiStatus StatusToScsi(nvme_defs::MediaErrorStatusCode sc) {
```
oicirtap, line 160: Same here

lib/translator/status.cc
```diff
@@ -0,0 +1,248 @@
...
+    default:
+      DebugLog("No SCSI translation for NVMe status with code %#x",
+               static_cast<uint8_t>(sc));
+  }
+  return result;
+}
+
+}  // namespace
+
+ScsiStatus StatusToScsi(uint8_t sct, uint8_t sc) {
```
oicirtap, line 226: same here, both for `sct`, and `sc`

PR #63, opened 2020-06-15
Patching master Makefile

basimsahaf:


harloff-g:
At this point in development does it make sense to just remove the hello world files?

basimsahaf:
Yeah I think it sounds reasonable to do that at this point

Issue #59, opened 2020-06-10
Implement a “Hello World” kernel module which mocks a SCSI device.


harloff-g:
We should have a proof of concept for emulating a SCSI device with a kernel module.

harloff-g:
Building scsi_debug as a kernel module & attaching the built module to the kernel produces promising results. The output of the command "lsscsi" is shown below, the mocked device being called "scsi_debug".

```
[0:0:1:0]    disk    Google   PersistentDisk   1     /dev/sda 
[1:0:0:0]    disk    Linux    scsi_debug       0188  /dev/sdb
```

basimsahaf:
I got the same result and the mock device is now showing up:

```
basimsahaf@basim-cloudtop:~/Desktop/workspace/scsi2nvme/modules$ lsscsi
[0:0:1:0]    disk    Google   PersistentDisk   1     /dev/sda 
[1:0:0:0]    disk    Linux    scsi_debug       0188  /dev/sdb 
```

harloff-g:
Status - Currently working on writing a "hello world" SCSI device mock driver from the ground up

PR #64, opened 2020-06-15
SCSI Mock Module

harloff-g:
Updates on #59 

This code provides an absolute bare-bones skeleton for mocking any type of SCSI device in the Linux kernel. It was built from the ground-up using various resources documenting Linux driver development.

The module successfully builds & attaches to the kernel.

Output is extremely promising, reading:

`[ 2872.630046] Registering root device`
`[ 2872.635006] Registering bus`
`[ 2872.639627] Registering mock driver`
`[ 2872.645553] Registering mock device`
`[ 2872.651184] Running device_register`
`[ 2872.657390] scsi host1: SCSI Mock Host, Version 0.1`
`[ 2872.664759] SUCCESS!`
`[ 2872.667762] RECIEVED COMMAND`
`[ 2894.216174] RECIEVED COMMAND`
`[ 2915.206778] RECIEVED COMMAND`
`[ 2936.197350] RECIEVED COMMAND`
`[ 2957.187841] RECIEVED COMMAND`
`[ 2978.178509] RECIEVED COMMAND`
`[ 2999.169118] RECIEVED COMMAND`

Currently, lsscsi lists no devices. This is because the module must respond to numerous SCSI commands (Mode Sense, Report Opcodes, Inquiry, Etc) to describe the device type. These commands can be responded to through the NVMe2SCSI translation engine, or directly from the module stub itself.

Any thoughts, comments, or concerns are welcomed & encouraged.


modules/scsi_mock/scsi_mock_module.c
```diff
@@ -0,0 +1,138 @@
...
+  device_unregister(&pseudo_adapter);
+  driver_unregister(&scsi_mock_driverfs);
+  bus_unregister(&pseudo_bus);
+  root_device_unregister(pseudo_root_dev);
+  printk("GOODBYE!\n");
+}
+
+device_initcall(scsi_mock_init);
+module_exit(scsi_mock_exit);
+MODULE_LICENSE("GPL");
```
Ralith, line 137: This code is licensed under Apache 2.0, not GPL.

harloff-g: Changing the license from GPL to any other license causes Kernel build to fail:
`GPL-incompatible module scsi_mock_module.ko uses GPL-only symbol 'root_device_unregister'`

This seems pretty nuanced, and I think the exact legal ramification of this may be out of our teams scope. Maybe we should consider consulting the open-source licensing team?

More context on this:
https://opensource.stackexchange.com/questions/4185/what-legal-meaning-do-gpl-symbols-hold-in-the-linux-kernel

Ralith: Hmm, interesting. Let's follow up internally, yeah.

Ralith: Conclusion is that module code should be GPL-licensed and must be moved into a subdir of `third_party` to help distinguish it from the Apache 2.0 library code.

modules/scsi_mock/scsi_mock_module.c
```diff
@@ -0,0 +1,138 @@
+#include <linux/device.h>
+#include <linux/init.h>
+#include <linux/kernel.h>
+#include <linux/module.h>
+#include <linux/scatterlist.h>
+
+#include <scsi/scsi.h>
+#include <scsi/scsi_host.h>
+
+#define NAME "SCSI2NVMe Mock"
```
w-michal, line 10: Prefer consts over macros. https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros (unless this is not possible in the module code)

modules/scsi_mock/scsi_mock_module.c
Ralith: Unresolving as this code still needs to be moved into third_party.

third_party/scsi_mock/scsi_mock_module.c
```diff
@@ -0,0 +1,137 @@
...
+
+static int scsi_queuecommand(struct Scsi_Host* host, struct scsi_cmnd* cmd) {
+  printk("RECIEVED COMMAND");
+  return 0;
+}
+
+static int scsi_abort(struct scsi_cmnd* cmd) { return SUCCESS; }
+
+static const char* scsi_mock_info(struct Scsi_Host* host) {
+  return "SCSI Mock Host, Version 0.1";
```
toadchild, line 29: Consider defining a macro for the module version so that you can use it here (via raw string concatenation) and in the registration macro at the end of the file.

harloff-g: Good point, that's definitely more maintainable! I've gone ahead and added that feature

harloff-g:
> third_party/scsi_mock should have a LICENSE file with a copy of the GPL.

Do you know if this license should be Google's GPL license or Linux's GPL license?

third_party/scsi_mock/scsi_mock_module.h
```diff
@@ -0,0 +1,28 @@
...
+#include <linux/device.h>
+
+static const char kName[] = "SCSI2NVMe SCSI Mock";
+static const int kQueueCount = 1;
+
+static struct bus_type pseudo_bus;
+static struct device* pseudo_root_dev;
+static struct device pseudo_adapter;
+static struct device_driver scsi_mock_driverfs = {.name = kName,
+                                                  .bus = &pseudo_bus};
```
Ralith, line 26: Is it idiomatic for these to be defined in the header?

harloff-g: After reviewing other similar code in the linux kernel it seems that this is atypical header use, so I'll move non-constant definitions to the .c file

third_party/scsi_mock/scsi_mock_module.h
```diff
@@ -0,0 +1,20 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+#ifndef NVME2SCSI_SCSI_MOCK_H
+#define NVME2SCSI_SCSI_MOCK_H
+
+#define VERSION "0.1"
+
+static const char kName[] = "SCSI2NVMe SCSI Mock";
+static const int kQueueCount = 1;
```
Ralith, line 18: I also wonder what the benefit of having even these in the header is--external code probably doesn't want a symbol called `kName` and we don't have multiple source files to share these between--but constants are much lower risk, at least.

PR #66, opened 2020-06-15
NVMe IO library

basimsahaf:
This PR contains code which sends commands directly to the NVMe device.


lib/nvme.h
```diff
@@ -0,0 +1,68 @@
...
+//    	__u32	cdw11;
+//    	__u32	cdw12;
+//    	__u32	cdw13;
+//    	__u32	cdw14;
+//    	__u32	cdw15;
+//    	__u32	timeout_ms;
+//    	__u32	result;
+//    };
+
+int send_passthru(int file_descriptor, unsigned long ioctl_command, uint8_t opcode, uint8_t flags, uint16_t rsvd1, uint32_t nsid,
```
w-michal, line 58: This function should accept `nvme_passthru_cmd` directly.

basimsahaf: this PR is actually old and contains code that talks to a device from userspace. Notice the file descriptors. Closing this now as we're no longer using it

PR #70, opened 2020-06-18
Translation Code Structure Proposal

JustinAWei:
This is a proposal + example of how to structure future translation code.
**tl;dr: each SCSI command has its own "main engine" `translate(raw, len)` logic**. 

Problem: Our previous approach tried to separate translation of all commands into distinct steps which was difficult due to huge variations in translation processes. Some processes had extra steps, parameters, etc. A simple yet major problem with a generic engine approach is the million types of structs we have--difficult to implement all of that control flow in a single generic engine.

Solution: "main engines" for each command to allow for command-level flexibility, and allows us to assume some pre/post conditions (e.g we can assume we are operating on a certain command). This approach still allows modularity, e.g all nvme execution functions can be part of a library apart from the command-specifc engine. This is really simply a way to shift control flow away from a central main engine for every command.

Command specific main logic would look like this (identical to previous generic engine)
1. get raw intercepted scsi command as input
2. verify and cast to a command specific struct (e.g `raw_cmd_to_scsi_command()`)
3. examine command specific parameters (e.g. `evpd` and `page_code` for inquiry) for control flow
4. call more specific functions (including nvme) to build/execute these results. e.g `build_standard_inquiry_data()` and `nvme_identify_namespace()`

Control flow from the interceptor will look like this. Allows us to have only one place where we identify the command.
```
intercept_module() {
    uint8_t *cmd = intercept();
    uint8_t opcode = cmd[0];
    switch (opcode) {
        case 0x12: inquiry::translate(cmd, len);
        case ...: write::translate(cmd, len);
        case ...: read::translate(cmd, len);
        ....
    }
}
```

Please let me know your thoughts on this approach.

### DIFF
--- a/lib/translate/inquiry.cc
+++ b/lib/translate/inquiry.cc
@@ -38,6 +38,7 @@ namespace inquiry {
         // TODO
         return nvme_defs::IdentifyControllerData();
     }
+
     // Executes the NVME Identify Namespace command
     nvme_defs::IdentifyNamespace nvme_identify_namespace() {
         // TODO
@@ -74,20 +75,17 @@ namespace inquiry {
         strncpy(result.vendor_identification, NVME_VENDOR_IDENTIFICATION, 8);
 
         // Shall be set to the first 16 bytes of the Model Number (MN) field within the Identify Controller Data Structure
-        for(int i = 0; i < 16; i++) {
-            result.product_identification[i] = identify_controller_data.mn[i];
-        }
+        memcpy(result.product_identification, identify_controller_data.mn, 16);
 
-        /* Shall be set to the last 4 ASCII graphic characters in the range of 21h-7Eh (i.e. last 4 non-space
+        /* 
+        Shall be set to the last 4 ASCII graphic characters in the range of 21h-7Eh (i.e. last 4 non-space
         characters) of the Firmware Revision (FR) field within the Identify Controller Data Structure.
         */
         int idx = 3;
         for(int i = 7; i >= 0; i--) {
             if(identify_controller_data.fr[i] != ' ' && identify_controller_data.fr[i] >= 0x21 && identify_controller_data.fr[i] <= 0x7e) {
                 result.product_revision_level[idx] = identify_controller_data.fr[i];
-                // printf("IDX %d\n", idx);
-                idx--;
-                if(idx == -1) break;
+                if(idx-- == 0) break;
             }
         }
         if(idx >= 0) printf("less than four characters set\n");
@@ -128,16 +126,13 @@ namespace inquiry {
         result.peripheral_qualifier = scsi_defs::PeripheralQualifier::kPeripheralDeviceConnected;
         result.peripheral_device_type = scsi_defs::PeripheralDeviceType::kDirectAccessBlock;
         result.page_code = 0x80;
-
         
         // check if nonzero
         bool nguid_nz = identify_namespace_data.nguid[0] || identify_namespace_data.nguid[1];
         bool eui64_nz = identify_namespace_data.eui64;
-        printf("e %d n %d\n", eui64_nz, nguid_nz);
         if(eui64_nz) {
             if(nguid_nz) {
                 // 6.1.3.1.1
-                // nguid
                 result.page_length = 40;
                 char hex_string[33];
                 sprintf(hex_string, "%llx", identify_namespace_data.nguid[0]);
@@ -161,9 +156,7 @@ namespace inquiry {
             }
             else {
                 // 6.1.3.1.2
-                // eui64
                 result.page_length = 20;
-                // printf("%llx\n", identify_namespace_data.eui64);
 
                 char hex_string[17];
                 sprintf(hex_string, "%llx", identify_namespace_data.eui64);
--- a/lib/translate/inquiry.cc
+++ b/lib/translate/inquiry.cc
@@ -131,19 +131,42 @@ namespace inquiry {
 
         
         // check if nonzero
-        bool eui64_nz = false, nguid_nz = false;
-        for(int i = 0; i < 16; i++) {
-            if(identify_namespace_data.nguid[i] != 0) {
-                nguid_nz = true; break;
-            }
-        }
-        eui64_nz = identify_namespace_data.eui64 != 0;
+        bool nguid_nz = identify_namespace_data.nguid[0] || identify_namespace_data.nguid[1];
+        bool eui64_nz = identify_namespace_data.eui64;
         printf("e %d n %d\n", eui64_nz, nguid_nz);
         if(eui64_nz) {
             if(nguid_nz) {
                 // 6.1.3.1.1
                 // nguid
                 result.page_length = 40;
+                // char hex_string[17];
+                // sprintf(hex_string, "%llx", identify_namespace_data.nguid[0]);
+
+                // char formatted_hex_string[21];
+                // for(int i = 4; i < result.page_length; i+=5) {
+                //     formatted_hex_string[i] = '_';
+                // }
+
+                // int pos = 0;
+                // for(int i = 0; i < result.page_length - 1; i++) {
+                //     if(formatted_hex_string[i] != '_') {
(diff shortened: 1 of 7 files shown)


lib/translate/inquiry.cc
```diff
@@ -0,0 +1,154 @@
...
+        result.supported_page_list[3] = 0x86;
+        result.supported_page_list[4] = 0xB0;
+        result.supported_page_list[5] = 0xB1;
+        result.supported_page_list[6] = 0xB2;
+        return result;
+    }
+
+    // TODO: figure out return value...
+    // Main logic engine for the Inquiry command
+    void translate(uint32_t* raw_cmd, int buffer_length) {
```
Ralith, line 120: This should take an `absl::Span<const uint32_t>`, assuming `buffer_length` is a number of `uint32_t`s.

JustinAWei:
@Ralith For the public interface, would it be sufficient to simply expose one function that takes in raw scsi data and spits out a nvme result (with extras like errors)? e.g a simple `translate()` containing the `switch(opcode)` logic in the intercept module. If it does need to be more complex than that I will look into making a design doc for the public facing side. 

I will also keep in mind translations that need state.

Ralith:
I think that's a great plan. If/when we need more, it should become apparent as e.g. the kernel module work proceeds, and should be able to build gracefully on top of that.

harloff-g:
> @Ralith For the public interface, would it be sufficient to simply expose one function that takes in raw scsi data and spits out a nvme result (with extras like errors)? e.g a simple `translate()` containing the `switch(opcode)` logic in the intercept module. If it does need to be more complex than that I will look into making a design doc for the public facing side.
> 
> I will also keep in mind translations that need state.

How will we facilitate two-pass memory allocation in the engine with this interface?

Ralith:
> How will we facilitate two-pass memory allocation in the engine with this interface?

It might be easier to consider a specific example case where this is necessary, but in general, we'll be writing NVMe commands/SCSI responses to caller-supplied output buffers as mandated by the SCSI/NVMe specs, and making multiple passes over input to if it's necessary to do so.

harloff-g:
> It might be easier to consider a specific example case where this is necessary, but in general, we'll be writing NVMe commands/SCSI responses to caller-supplied output buffers as mandated by the SCSI/NVMe specs, and making multiple passes over input to if it's necessary to do so.

Gotcha. I was confusing this method as also translating NVMe responses but I've cleared up confusions with Justin :+1:

JustinAWei:
@harloff-g so each `inquiry::translate(cmd, len);` is it's own mini-engine which does end-to-end translation. See my file changes for an example. This means it parses the scsi parameters, calls the appropriate NVMe commands, and uses responses to build scsi results. The two-pass method could be used inside some logic after getting NVMe responses within the mini-engine.

eyhhuang:
I'm happy to have each mini engine do its own end to end translation!

harloff-g:
> @harloff-g so each `inquiry::translate(cmd, len);` is it's own mini-engine which does end-to-end translation. This means it parses the scsi parameters, calls the appropriate NVMe commands, and uses responses to build scsi results. The two-pass method could be used inside some logic after getting NVMe responses within the mini-engine.
> 
> See my file changes for an example. In `translate()` I call `build_standard_inquiry_data()` which starts handling a specific parameter combination within the identify command. It in turn calls `nvme_identify_controller()` and `nvme_identify_namespace()` which are NVMe commands used to build a scsi response struct. The two pass method could be used at this point to initialize the scsi response.

This seems good, but I will point out that the current code demonstrating this structure doesn't appear to offload any translation work to a translation library like we're primarily focused on creating. Also, nvme commands should probably be called through only one generic function, and not command specific functions, as the translation library is designed to return generic nvme commands.

basimsahaf:
Having a separate engine for each command is great design. I agree with @harloff-g that we need a common interface for all commands and not call each nvme command through specific functions.

JustinAWei:
@basimsahaf @harloff-g We can definitely have a separate generic NVMe communication library since NVMe communication is pretty separate from the command-specific engine.

JustinAWei:
@harloff-g I think it's simply a matter of how we want to organize the logic. My currently design is a function for each parameter-combo. e.g `build_standard_inquiry_data()` So calls to NVMe and translation are both inside that function. We can definitely separate all NVMe communication logic from translation by simply wrapping them into functions. 

@basimsahaf @eyhhuang Please let me know your thoughts on which direction we want to take.

JustinAWei:
Here is an example of how we can do this
```
scsi_defs::InquiryData build_standard_inquiry_data() {
        // NVMe calls
        nvme_defs::IdentifyControllerData identify_controller_data = nvme_identify_controller();
        nvme_defs::IdentifyNamespace identify_namespace_data = nvme_identify_namespace();
        
        // Translation of response
        scsi_defs::InquiryData result = scsi_defs::InquiryData();
        result.peripheral_qualifier = static_cast<scsi_defs::PeripheralQualifier>(0);
        result.peripheral_device_type = static_cast<scsi_defs::PeripheralDeviceType>(0);
        result.rmb = 0;
        ...
}
```
scsi_defs::InquiryData build_standard_inquiry_data() {
        // NVMe calls
        nvme_defs::IdentifyControllerData identify_controller_data = nvme_identify_controller();
        nvme_defs::IdentifyNamespace identify_namespace_data = nvme_identify_namespace();
        
        // Translation of response
        return translate_response(identify_controller_data, identify_namespace_data);
}

scsi_defs::InquiryData translate_response(...) {
        scsi_defs::InquiryData result = scsi_defs::InquiryData();
        result.peripheral_qualifier = static_cast<scsi_defs::PeripheralQualifier>(0);
        result.peripheral_device_type = static_cast<scsi_defs::PeripheralDeviceType>(0);
        result.rmb = 0;
        ...
}
```

File: lib/translate/inquiry.cc [hunk 2]
```diff
@@ -0,0 +1,200 @@
...
+            return std::nullopt;
+        }
+
+        // printf("opcode is good, casting\n");
+        // TODO: check invalid parameters in scsi_command
+        return std::make_optional(*reinterpret_cast<const scsi_defs::InquiryCommand*>(&raw_cmd[1]));
+    }
+
+    // Executes the NVME Identify Controller command
+    nvme_defs::IdentifyControllerData nvme_identify_controller() {
```
harloff-g, line 37: AFAIK lib/translate is intended to strictly act as a land for code to live which translates between SCSI and NVMe. A lot of functionality picked up in this file, including activities such as executing NVMe commands, renders the translation library reliant on downstream dependencies. The mini-engines are probably better placed in kernel module code at /modules. I attached a flow-chart to the Google doc which details an overview of how this code could be segmented

harloff-g: Alternatively, we could create a lib/engine directory for c++ engine code to live

eyhhuang: Just to clarify, according to the flow-chart, the code inside lib/translate should just be responsible for 
1) translating SCSI -> NVMe commands
2) translating NVMe -> SCSI responses
3) 2 pass memory translation

Everything else, including the parsing from raw bytes to scsi commands would be doone in the mini engine, which exists as a kernel module? 

I also agree that the translation lib should have the single responsibility of translation

harloff-g: The three core responsibilities of lib/translation that you outlined are the intention behind the flowchart, yeah. Just trying to keep different parts of the code organized with a clear and manageable scope.

There's still some unhandled cases which the chart doesn't explicitly deal with, like, as you mentioned, taking SCSI command buffers and casting them to our custom structures. I think writing a library for this task to be accessed from the engine/kernel module and thus enjoying the benefits of Bazel testing is a perfectly reasonable thing to do.

Things that should probably exist as inseparable from the engine are processes like memory allocation, as that is inherently tied to our full-scope implementation.

JustinAWei: That's a good idea-putting the modules that the engine deals with in other scopes / files. I will look into reorganizing.

harloff-g:
I think it's reasonable to close this PR given our current progress. Thoughts?

PR #72, opened 2020-06-22
Translate Report LUN

harloff-g:
Closes #72 

Code to translate the SCSI Report Luns command. In correspondence with the latest API design


third_party/spdk_defs/nvme_defs.h
```diff
@@ -328,6 +330,288 @@ struct GenericQueueEntryCmd {
 } ABSL_ATTRIBUTE_PACKED;
 static_assert(sizeof(GenericQueueEntryCmd) == 64);
 
+// NVMe Base Specification Figure 70 to Figure 75
```
harloff-g, line 333: I'm not sure why it shows 333-614 as new. They currently exist in master

third_party/spdk_defs/nvme_defs.h
```diff
@@ -452,6 +736,11 @@ struct IdentifyNamespace {
 } ABSL_ATTRIBUTE_PACKED;
 static_assert(sizeof(IdentifyNamespace) == 4096);
 
+struct IdentifyNamespaceList {
```
harloff-g, line 739: This structure is needed to handle the NVMe Identify command response

third_party/spdk_defs/nvme_defs.h
```diff
@@ -303,7 +305,7 @@ static_assert(sizeof(SglDescriptor) == 16);
 // https://nvmexpress.org/wp-content/uploads/NVM-Express-1_4-2019.06.10-Ratified.pdf
 struct GenericQueueEntryCmd {
   // dword 0
-  uint16_t opc : 8;   // opcode
+  uint8_t opc : 8;    // opcode
```
harloff-g, line 308: This field was altered to allow for warning-less opcode casting (opcode is stored as uint8_t).

lib/translate/common.h
```diff
@@ -15,10 +15,16 @@
 #ifndef LIB_TRANSLATOR_COMMON_H
 #define LIB_TRANSLATOR_COMMON_H
 
+namespace translator {
+
+enum class TranslatorStatusCode { SUCCESS, FAILURE };
```
Ralith, line 20:
```suggestion
enum class StatusCode { SUCCESS, FAILURE };
```

Avoid repeating the namespace in the name; downstream code that wants/needs to be explicit can qualify with the namespace as desired.

lib/translate/lun_translator.h
```diff
@@ -0,0 +1,37 @@
...
+#include <optional>
+
+#include "../../third_party/spdk_defs/nvme_defs.h"
+#include "../scsi_defs.h"
+#include "absl/types/span.h"
+#include "common.h"
+
+namespace translator {
+
+std::optional<nvme_defs::GenericQueueEntryCmd> translateReportLunCommand(
```
Ralith, line 27: Functions (here and below) should be named consistent with [the relevant section of the style guide](https://google.github.io/styleguide/cppguide.html#Function_Names).

lib/translate/lun_translator.cc
```diff
@@ -0,0 +1,70 @@
...
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+std::optional<nvme_defs::GenericQueueEntryCmd> translateReportLunCommand(
+    scsi_defs::ReportLunsCommand* scsi_cmd, uint64_t prp_ptr) {
+  if (scsi_cmd->select_report != scsi_defs::SelectReport::kRestrictedMethods &&
+      scsi_cmd->select_report != scsi_defs::SelectReport::kWellKnown &&
+      scsi_cmd->select_report != scsi_defs::SelectReport::kAllLogical)
+    return std::nullopt;
+
+  nvme_defs::GenericQueueEntryCmd nvme_cmd = {};
+  nvme_cmd.opc = (uint8_t)nvme_defs::AdminOpcode::kIdentify;
```
Ralith, line 29: [Use C++ casts](https://google.github.io/styleguide/cppguide.html#Casting). In this case, a `static_cast` would be appropriate.

third_party/spdk_defs/nvme_defs.h
Ralith: That won't be necessary if `static_cast` is used.

lib/translate/common.h
eyhhuang: I'll also make this change in https://github.com/googleinterns/scsi2nvme/pull/62

eyhhuang, line 20: Also, i think enum names should the kName format https://google.github.io/styleguide/cppguide.html#Enumerator_Names

harloff-g: Ahh, thanks for pointing that out, totally forgot!

lib/translate/lun_translator.h
```diff
@@ -0,0 +1,37 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIB_LUN_TRANSLATOR_H
+#define LIB_LUN_TRANSLATOR_H
+
+#include <optional>
+
+#include "../../third_party/spdk_defs/nvme_defs.h"
```
eyhhuang, line 20: Just realized that the style guide states `All of a project's header files should be listed as descendants of the project's source directory without use of UNIX directory aliases . (the current directory) or .. (the parent directory).`

test/BUILD
```diff
@@ -1,13 +1,17 @@
...
-        "//lib/translate:translator_lib",
-        "@googletest//:gtest_main",
-        "@com_google_absl//absl/base",
-    ],
+  name = "tests",
+  srcs = [
+    "scsi_defs_test.cc",
+    "nvme_defs_test.cc",
+    "status_translator_test.cc",
+    "lun_translator_test.cc"
```
eyhhuang, line 7: Would be good to put this in its own target. There's already a PR out that does so for master

lib/translator/report_luns.cc
```diff
@@ -0,0 +1,70 @@
...
+                                      absl::Span<uint8_t> span) {
+  if (span.size() < sizeof(scsi_defs::ReportLunsParamData))
+    return StatusCode::kFailure;
+
+  scsi_defs::ReportLunsParamData rlpd = {};
+  rlpd.list_byte_length = span.size() - sizeof(scsi_defs::ReportLunsParamData);
+  uint32_t lun_count = rlpd.list_byte_length / sizeof(scsi_defs::LunAddress);
+
+  if (lun_count > nvme_defs::kIdentifyNsListMaxLength)
+    return StatusCode::kFailure;
```
JustinAWei, line 58: print error msg

lib/translator/report_luns.cc
```diff
@@ -0,0 +1,70 @@
...
+StatusCode TranslateReportLunCommand(scsi_defs::ReportLunsCommand& scsi_cmd,
+                                     nvme_defs::GenericQueueEntryCmd& nvme_cmd,
+                                     uint64_t prp_ptr) {
+  if (scsi_cmd.select_report != scsi_defs::SelectReport::kRestrictedMethods &&
+      scsi_cmd.select_report != scsi_defs::SelectReport::kWellKnown &&
+      scsi_cmd.select_report != scsi_defs::SelectReport::kAllLogical)
+    return StatusCode::kFailure;
+
+  nvme_cmd = {};
+  nvme_cmd.opc = static_cast<uint8_t>(nvme_defs::AdminOpcode::kIdentify);
```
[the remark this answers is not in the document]
harloff-g: Huh, good idea! We might want to make a separate PR doing so

lib/translator/report_luns.cc
```diff
@@ -0,0 +1,70 @@
...
+    if (ns_list.ids[i] == 0) break;
+    ++size;
+  }
+  return size;
+}
+
+// Section 6.6
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+StatusCode TranslateReportLunResponse(nvme_defs::IdentifyNamespaceList& ns_list,
+                                      absl::Span<uint8_t> span) {
```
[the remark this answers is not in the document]
harloff-g: Span is intended to be a directly modifiable buffer here so it shouldn't be const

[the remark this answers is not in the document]
eyhhuang: do we want const on the IdentifyNamespace?

lib/translator/report_luns.h
```diff
@@ -0,0 +1,34 @@
...
+#include "common.h"
+#include "lib/scsi_defs.h"
+#include "third_party/spdk_defs/nvme_defs.h"
+
+namespace translator {
+
+StatusCode ReportLunsToNvme(absl::Span<const uint8_t> scsi_cmd,
+                            nvme_defs::GenericQueueEntryCmd& nvme_cmd);
+
+StatusCode ReportLunsToScsi(nvme_defs::GenericQueueEntryCmd,
```
Ralith, line 29: The definition of this takes a const reference, but this prototype takes a value. That mismatch will lead to linker errors.

harloff-g: Thanks!

harloff-g:
Refactoring to use interface provided by #104

lib/translator/report_luns.cc
```diff
@@ -0,0 +1,111 @@
...
+    return StatusCode::kInvalidInput;
+  }
+
+  // Assign allocation length for downstream use
+  alloc_len = rl_cmd.alloc_length;
+
+  // Assign generic params to NVMe command
+  identify_cmd = {};
+  identify_cmd.opc = static_cast<uint8_t>(nvme_defs::AdminOpcode::kIdentify);
+  identify_cmd.cdw[0] = 0x2;  // Set CNS to return namespace ID list
```
eyhhuang, line 61: for a CNS value of 0x2, I think the needs to be set. Figure 244 https://nvmexpress.org/wp-content/uploads/NVM-Express-1_4-2019.06.10-Ratified.pdf

harloff-g: In this case the NSID should be 0. Since the NVMe structure is 0 initialized when identify_cmd = {}, I neglected setting it redundantly. If the general consensus is to explicitly set the NSID to 0 we can do that, but imo it's cleaner to just 0 initialize

test/translator/report_luns_test.cc
```diff
@@ -0,0 +1,99 @@
...
+    EXPECT_EQ(i, lun_list[i]);
+  }
+}
+
+TEST(reportLunsToScsi, shouldFailNotEnoughMemory) {
+  nvme_defs::GenericQueueEntryCmd cmd_data;
+  uint8_t* buf = nullptr;
+  absl::Span<uint8_t> span = absl::MakeSpan(buf, 0);
+
+  translator::StatusCode actualStatus =
```
eyhhuang, line 93: nit: actual_status

harloff-g: Nice catch! So hard to let go of my camelcase ways lol

lib/translator/report_luns.cc
```diff
@@ -0,0 +1,110 @@
...
+}
+
+}  // namespace
+
+// Section 4.5
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+StatusCode ReportLunsToNvme(absl::Span<const uint8_t> scsi_cmd,
+                            nvme_defs::GenericQueueEntryCmd& identify_cmd,
+                            uint32_t& alloc_len) {
+  if (scsi_cmd.size() != sizeof(scsi_defs::ReportLunsCommand)) {
```
eyhhuang, line 40: In other PRs we've been delegating the size check to checking the return value of ReadValue() but keeping the same DebugLof("Malformed ReportLuns command")

lib/translator/report_luns.cc
```diff
@@ -0,0 +1,110 @@
...
+  // Assign allocation length for downstream use
+  alloc_len = rl_cmd.alloc_length;
+
+  // Assign generic params to NVMe command
+  identify_cmd = {};
+  identify_cmd.opc = static_cast<uint8_t>(nvme_defs::AdminOpcode::kIdentify);
+  identify_cmd.cdw[0] = 0x2;  // Set CNS to return namespace ID list
+
+  // Allocate prp & assign to command
+  uint64_t prp_ptr = AllocPages(1);
```
eyhhuang, line 64: shameless plug to use https://github.com/googleinterns/scsi2nvme/pull/116

lib/translator/report_luns.cc
```diff
@@ -0,0 +1,110 @@
...
+  uint8_t* dptr = reinterpret_cast<uint8_t*>(identify_cmd.dptr.prp.prp1);
+  ReadValue(absl::MakeSpan(dptr, sizeof(ns_list)), ns_list);
+  uint32_t lun_count = GetNsListLength(ns_list);
+
+  // Define scsi response structure and determine attributes
+  uint32_t allocated_list_bytes =
+      buffer.size() - sizeof(scsi_defs::ReportLunsParamData);
+  scsi_defs::ReportLunsParamData rlpd = {};
+  rlpd.list_byte_length = lun_count * sizeof(scsi_defs::LunAddress);
+  if (rlpd.list_byte_length > allocated_list_bytes) {
```
toadchild, line 90: Log an error if you're truncating the returned list of LUNs

lib/translator/report_luns.cc
```diff
@@ -0,0 +1,110 @@
...
+        (allocated_list_bytes % sizeof(scsi_defs::LunAddress));
+    lun_count = rlpd.list_byte_length / sizeof(scsi_defs::LunAddress);
+  }
+
+  // Write response to buffer
+  WriteValue(rlpd, buffer);
+  uint32_t lun_offset = sizeof(scsi_defs::ReportLunsParamData);
+  for (uint32_t i = 0; i < lun_count; ++i) {
+    scsi_defs::LunAddress lun = static_cast<scsi_defs::LunAddress>(
+        ns_list.ids[i] - 1);  // LUN must start @ 0 via SAM spec
```
toadchild, line 102: Are other parts of the project consistently applying this same transformation?  Do we want a shared helper function to formalize mapping a LUN to an NSID and back?

harloff-g: Great point! Currently other parts of the project would have to apply this same transformation, but I think it's reasonable to transform the lun to nsid in the "begin" function, and then downstream code can use it without knowing translation details

lib/translator/report_luns.cc
```diff
@@ -0,0 +1,111 @@
...
+  if (rlpd.list_byte_length > allocated_list_bytes) {
+    rlpd.list_byte_length = allocated_list_bytes -
+                            (allocated_list_bytes % sizeof(scsi::LunAddress));
+    lun_count = rlpd.list_byte_length / sizeof(scsi::LunAddress);
+  }
+
+  // Write response to buffer
+  if (!WriteValue(rlpd, buffer)) {
+    DebugLog("Buffer not large enough for report luns response header");
+    return StatusCode::kSuccess;
```
eyhhuang, line 95: why do we return success here?

harloff-g: Truncating variable length data in SCSI is not meant to be an error

lib/translator/report_luns.cc
```diff
@@ -0,0 +1,125 @@
...
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+StatusCode ReportLunsToScsi(const nvme::GenericQueueEntryCmd& identify_cmd,
+                            Span<uint8_t> buffer) {
+  if (buffer.size() < sizeof(scsi::ReportLunsParamData)) {
+    DebugLog("Insufficient buffer size");
+    return StatusCode::kFailure;
+  }
+
+  // Get ns_list and lun count from NVMe dptr
+  uint8_t* ns_dptr = reinterpret_cast<uint8_t*>(identify_cmd.dptr.prp.prp1);
```
JustinAWei, line 84: is safe ptr meant to be used here?

test/translator/report_luns_test.cc
```diff
@@ -0,0 +1,114 @@
...
+// limitations under the License.
+
+#include "lib/translator/report_luns.h"
+#include "lib/translator/common.h"
+
+#include "absl/base/casts.h"
+#include "gtest/gtest.h"
+
+#include <byteswap.h>
+#include <netinet/in.h>
```
JustinAWei, line 22: do we need another kernel check here?

lib/translator/report_luns.cc
harloff-g: Unfortunately the buffer pointer `identify_cmd.dptr.prp.prp1` is of type uint64_t and SafePointerCastRead expects a span of type uint8_t. What's happening here is we are creating that span which is used in line 87

test/translator/report_luns_test.cc
harloff-g: The Bazel testing suite runs in user space & thus these tests are not runnable in the kernel, so I omitted the kernel check

Issue #74, opened 2020-06-22
Write an optional template that works in the kernel space


eyhhuang:
As @harloff-g pointed out in https://github.com/googleinterns/scsi2nvme/issues/71, the std::optional results in some linking issues when used in the kernel space. As the optional class offers some useful features, notably to clearly indicate when a value has been initialized, we will create our own template class that implements the necessary features

eyhhuang:
See https://github.com/googleinterns/scsi2nvme/pull/75#issuecomment-649020238 for resolution details.

PR #75, opened 2020-06-23
Optional template

eyhhuang:
Closes https://github.com/googleinterns/scsi2nvme/issues/74

See SCSI to NVMe Custom Structs design doc for more comments


lib/optional.h
```diff
@@ -0,0 +1,20 @@
+#ifndef LIB_OPTIONAL_H
```
harloff-g, line 1: nit: could be good for organization to put util structures in a subdirectory

lib/optional.h
```diff
@@ -0,0 +1,20 @@
...
+
+namespace util {
+
+template <typename T>
+struct Optional {
+  bool isNull;
+  T data;
+
+  Optional() : isNull{true} {}
+  Optional(const T& data) : isNull{false}, data{data} {}
```
harloff-g, line 12: consider overriding the = operator so that when the optional is set equal to something it also is set to isNull=false

eyhhuang: good idea

eyhhuang: Actually im not sure if we need to overload = for copy/move assignment. If the Optional is set equal to something, we'll need to check whether that something has a non null field before setting isNull = false. And I think the behavior ends up being the same as the default bytewise copy.

harloff-g: In the case that we set an optional equal to another optional I totally agree. However, it may be the case that we wish to set the optional data with just Optional<type> op = type, which is the case I'm referring to. Do you think this case could be useful to cover?

eyhhuang: Yeah I think that would be useful. Didn't know that you could overload = with another type!

lib/util/optional.h
```diff
@@ -0,0 +1,21 @@
+#ifndef LIB_UTIL_OPTIONAL_H
+#define LIB_UTIL_OPTIONAL_H
+
+namespace util {
+
+template <typename T>
+struct Optional {
```
Ralith, line 7: Consider defining `explicit operator bool()` for convenience of state checking.

lib/util/optional.h
```diff
@@ -0,0 +1,21 @@
...
+  bool isNull;
+  T data;
```
Ralith, line 9: These members should be private to help make access to possibly-unset values more obvious through the use of accessors.

lib/util/optional.h
```diff
@@ -0,0 +1,21 @@
...
+
+  Optional() : isNull{true} {}
+  Optional(const T& data) : isNull{false}, data{data} {}
+
+  bool isEmpty() const { return this->isNull; }
+  void setEmpty() { this->isNull = true; }
+  T get() const { return this->data; }
```
Ralith, line 16: Prefer returning a const reference here.

eyhhuang:
> Should we consider pursuing something that encodes a status code in the empty case, given the motivation to use this for fallible functions?
> 
> Note that as written, this is only useful for types that are default-constructible and cheap to copy. I think that describes all our expected use of it, so that's fine, but that might give some insight into why `std::optional` is much more complicated.

Yeah the need of default ctors and the potential performance impacts are definitely an issue. For the `TranslatedScsiStatus`, it's not too big of problem, but we were also considering using the optional to wrap larger structs as in here https://github.com/googleinterns/scsi2nvme/pull/70/files#diff-ef79bf8d5306904c52d657829b560ef3R19

I'll dig around for some other potential solutions

eyhhuang:
After some discussion with the other interns, we decided to not use the custom optional class due to concerns regarding the necessity and performance of the default constructor. Instead, we will use the pattern of passing the data by reference and modifying it in the function, which then returns a status code. This is also consistent with the plan to handle variable-sized data. The branch will be left undeleted for a bit just in case it becomes relevant.

Ralith:
> we were also considering using the optional to wrap larger structs as in here

By "cheap to copy" I mostly meant "have a trivial copy constructor," i.e. no implicit heap allocation or similar. In general I'd trust the optimizer to handle passing data around by-value sensibly even if a naive compilation would have a lot of redundant work.

Standardizing on returning a status code + possibly writing to an output parameter is fine too, though, especially since that'll likely let us write directly to the actual shared memory buffers, which is good fun.

PR #77, opened 2020-06-25
Define common translation definitions

harloff-g:
Defines common status codes and logging methods for the translation library


lib/translate/common.h
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace translator {
+
+enum class StatusCode { kSuccess, kInvalidInput, kNoTranslation, kFailure };
+
+void DebugLog(const char* buffer);
```
Ralith, line 19: Consider transforming this into a C-style variadic function which internally invokes `vsnprintf`, for convenience of performing nontrivial formatting.

test/translate/common_test.cc
```diff
@@ -0,0 +1,36 @@
...
+
+namespace {
+
+/*
+   Tests the logging methods
+*/
+
+const char* actual_buf = nullptr;
+
+void callback(const char* buf) { actual_buf = buf; }
```
Ralith, line 27: These should be defined inside the scope of the test. While a function definition like `callback` can't exist there, you can instead use a lambda expression with an empty capture list, e.g. `[](const char* buf) { ... }`. You could use a static local defined in the test scope from inside the lambda, but better yet would be to just place the `EXPECT_EQ` directly inside the lambda.

PR #86, opened 2020-06-29
Test cpp20

JustinAWei:


harloff-g:
It appears that the build error is not caused by c++2a so I'm closing this PR for now

PR #87, opened 2020-06-29
Translation Library Public Interface

harloff-g:
The primary intent of this PR is to define all public-facing translation library structures and classes.

A rough guideline implementation of the translation object is provided. Specific error handling & flow is not currently well defined, but the current framework afforded by the state of the translation object seems to allow for all error handling I could forsee.

The kInquiry case is an example of how control-flow will be delegated to downstream functions, and it can be updated as required in the Inquiry translation PR.


lib/translator/translation.cc
```diff
@@ -0,0 +1,73 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "common.h"
+
+namespace translator {
+
+StatusCode Translation::ScsiToNvme(absl::Span<const uint8_t> scsi_cmd,
```
harloff-g, line 19: Would it be more idiomatic for the functionality of this method to take place in the class constructor, or is this fine?
Currently, interaction with the library expects a strict adherence to ScsiToNvme->GetNvmeCmds->NvmeToScsi. Is this fine from an idiomatic standpoint @Ralith ?

lib/translator/common.h
```diff
@@ -23,6 +28,34 @@ void DebugLog(const char* format, ...);
...
+
+void SetAllocPageCallback(void* (*callback)(uint8_t));
+
+struct ScsiContext {
+  scsi_defs::LunAddress lun;
+};
+
+struct NvmeCompletionData {
+  nvme_defs::GenericQueueEntryCpl nvme_cpl[3];
+};
```
Ralith, line 41: Do we have strong reason to expect these will need multiple fields in the future? If not, the extra type seems like unnecessary boilerplate.

lib/translator/common.h
```diff
@@ -23,6 +28,34 @@ void DebugLog(const char* format, ...);
...
+void* AllocPage(uint8_t count);
+
+void SetAllocPageCallback(void* (*callback)(uint8_t));
+
+struct ScsiContext {
+  scsi_defs::LunAddress lun;
+};
+
+struct NvmeCompletionData {
+  nvme_defs::GenericQueueEntryCpl nvme_cpl[3];
```
Ralith, line 40: Could this be an `absl::Span<const nvme_defs::GenericQueueEntryCpl>` instead?

lib/translator/translation.cc
Ralith: Constructors cannot fail except by throwing an exception, which isn't usually appropriate in general and is specifically forbidden by the Google style guide. An alternative would be a static method which returns a new object or an error, but because we don't have convenient access to templates that make that easy, the approach you currently have is much simpler.

Additionally, downstream code might want to keep a preallocated table of these structures and rotate through them, in which case repeatedly iterating that cycle on a long-lived object is convenient.

lib/translator/translation.cc
```diff
@@ -0,0 +1,73 @@
...
+                                   void* data_in) {
+  if (this->pipeline_status == StatusCode::kFailure) {
+    // TODO fill data-in buffer with SCSI failure response
+    return StatusCode::kFailure;
+  }
+
+  if (scsi_cmd.size() < 1) {
+    DebugLog("Empty SCSI Buffer");
+    return StatusCode::kFailure;
+  }
```
Ralith, line 53: Is this intentionally duplicating the check performed above in `ScsiToNvme`?

lib/translator/common.h
```diff
@@ -23,6 +28,34 @@ void DebugLog(const char* format, ...);
...
+};
+
+class Translation {
+ public:
+  Translation() = default;
+  StatusCode ScsiToNvme(absl::Span<const uint8_t> scsi_cmd,
+                        const ScsiContext& scsi_context);
+  StatusCode NvmeToScsi(const NvmeCompletionData& cpl_data, void* data_in);
```
Ralith, line 48: nit: I feel like uses of these might be more readable (i.e. less dyslexia-inducing) if these were renamed `Start`/`Finish` or `Begin`/`End` or something along those lines.

lib/translator/translation.cc
```diff
@@ -0,0 +1,73 @@
...
+namespace translator {
+
+StatusCode Translation::ScsiToNvme(absl::Span<const uint8_t> scsi_cmd,
+                                   const ScsiContext& scsi_context) {
+  this->pipeline_status = StatusCode::kSuccess;
+  // Verify buffer is large enough to contain opcode (one byte)
+  if (scsi_cmd.size() < 1) {
+    DebugLog("Empty SCSI Buffer");
+    this->pipeline_status = StatusCode::kFailure;
+    return StatusCode::kFailure;
```
Ralith, line 26: Are there cases where the caller needs to respond differently to different status codes from this function, or should it always unconditionally proceed to execute commands (if any) and finish with `NvmeToScsi`? If the latter, are we actually getting any value from returning an error code here?

It might be useful to consider whether we have both, and can isolate, cases of "the proximate caller is misusing this API" and "the caller is, through no fault of its own, passing on malformed SCSI data it received and needs us to come up with some sort of sensible response to." For the latter case the caller might not need to care whether the SCSI operation is well-formed.

lib/translator/common.h
```diff
@@ -23,6 +28,34 @@ void DebugLog(const char* format, ...);
...
+  scsi_defs::LunAddress lun;
+};
+
+struct NvmeCompletionData {
+  nvme_defs::GenericQueueEntryCpl nvme_cpl[3];
+};
+
+class Translation {
+ public:
+  Translation() = default;
```
Ralith, line 45: We probably don't need to have this here explicitly.

lib/translator/common.h
Ralith, line 48: If we're writing to `data_in`, we need *very* strong evidence that there's no danger of an arbitrarily malformed or maliciously crafted `scsi_cmd` causing us to write out of bounds. Furthermore, that evidence should be clearly cited here for the benefit of future readers.

Otherwise, we should replace this with an `absl::Span` and take care to bounds-check writes.

lib/translator/common.h
```diff
@@ -23,6 +28,34 @@ void DebugLog(const char* format, ...);
 
 void SetDebugCallback(void (*callback)(const char*));
 
+void* AllocPage(uint8_t count);
```
Ralith, line 31: Why `uint8_t`? Is that the maximum SCSI can possibly require? Maybe worth a comment explaining this.

lib/translator/translation.cc
```diff
@@ -0,0 +1,73 @@
...
+  this->scsi_cmd = scsi_cmd;
+  this->scsi_context = scsi_context;
+
+  scsi_defs::OpCode opc = static_cast<scsi_defs::OpCode>(scsi_cmd[0]);
+  switch (opc) {
+    case scsi_defs::OpCode::kInquiry:
+      return StatusCode::kSuccess;
+    default:
+      DebugLog("Bad OpCode");
+      this->pipeline_status = StatusCode::kFailure;
```
eyhhuang, line 38: I'm not too sure here when/how the pipeline_status differs from the StatusCode that's returned by the function.

harloff-g: Pipeline status allows the translation library to keep track of errors that have occurred in upstream interactions and act accordingly without breaking the interface.

For example, it could be the case that there was no translation possible for some given scsi_data. The translation object will set its internal state to NoTranslation. As the pipeline continues, consumers get the NVMe commands using GetNvmeCmds (which returns a span of size 0), and then consumers proceed to call NvmeToScsi.
NvmeToScsi will recognize that the earlier translation attempt failed and can then respond with appropriate error codes. Meanwhile, the consumer does not notice any change in the library interface.

lib/translator/translation.cc
harloff-g: This check was redundant assuming that the library is used properly. Removing

lib/translator/common.h
harloff-g: At what point should we bound check writes? I'm not super familiar with memory validation best practices.

Ralith:
> One primary question: How should we handle de-allocation of physical pages provided by the physical page callback? Another callback where the engine provides the pointer and the count?

Yeah, downstream code needs to be able to supply both. Probably even sensible to populate the globals with a single setter that takes two arguments, even, to express this clearly.

> The library will know exactly when it is done with the physical page data, however, this opens up room for bugs as we need to manually de-allocate.
>
> We could also keep track of all PRPs allocated by the translation object and de-allocate them in the translation object destructor.

We're guaranteed to be done with them after `NvmeToScsi` returns, right? If so, I'd favor freeing them there, as this simplifies reuse of the same object for future translations.

lib/translator/translation.cc
harloff-g: In theory the caller never actually has to respond to status codes and the library will still operate as intended. The main reason I was returning status codes was just to offer some insight to library consumers into the status of the SCSI commands the engine is handling.

I think your point about differentiating between API abuse and legitimate malformed data is really interesting. Do you have any insights into how something like this could be accomplished?

lib/translator/common.h
```diff
@@ -23,6 +28,34 @@ void DebugLog(const char* format, ...);
...
+  StatusCode Complete(
+      absl::Span<const nvme_defs::GenericQueueEntryCpl> cpl_data,
+      absl::Span<uint8_t> buffer);
+  // Returns a span containing translated NVMe commands.
+  absl::Span<const nvme_defs::GenericQueueEntryCmd> GetNvmeCmds();
+
+ private:
+  StatusCode pipeline_status;
+  absl::Span<const uint8_t> scsi_cmd;
+  uint32_t nvme_cmd_count;
```
eyhhuang, line 55: im not sure why we have this field but in the next line we have the size of nvme_cmds[] set to 3.

eyhhuang:
What happened to the ScsiContext struct that contained LunAddress? I was about to make a comment that that might have been a good place to store the prp_ptr

lib/translator/common.h
harloff-g: The number of valid NVMe cmds will not always be 3, 3 is just estimated as the maximum number we will ever need. Keeping this field allows us to keep track of how many valid commands we should return to the consumer when they call GetNvmeCmds

harloff-g:
> What happened to the ScsiContext struct that contained LunAddress? I was about to make a comment that that might have been a good place to store the prp_ptr

It was removed as at the moment there's nothing else we've identified to put there. prp_ptr will be allocated using the callback defined in common.h so we don't need it in scsi_context (won't be exposed to consumer)

lib/translator/common.h
Ralith: Any time you write to memory you need to prove that the memory is properly allocated and aligned for that write to be legal. Typically the type system does most of this work for you, but with `void*` all that goes out the window.

An `absl::Span` carries dynamic length and (implicit in its template argument) static alignment information, so each write to one should first check that the range of indices being written to fall within the supplied length, unless such a check can be easily proven redundant.

lib/translator/translation.cc
Ralith: If the behavior of the caller should be identical regardless of this method's return value, I'd favor omitting the return value entirely for clarity. Normally I'd expect a library interface returning an error to indicate that no further progress can be made.

> Do you have any insights into how something like this could be accomplished?

In this case, I think the structure of the problem lends itself to a very natural solution: bad SCSI commands should result in genuine SCSI error responses and successful (or entirely omitted) library API results, whereas misuse of the API, where possible, should yield an API-level error instead. It's not obvious to me that there's any need to intermix them to begin with.

In general, if you need to intermix different classes of errors, a good way to distinguish them is by exposing various classifier predicates that can be invoked on an error code. When practical, though, it's always much better to use distinct types or control flow to avoid any risk of confusion.

lib/translator/translation.cc
harloff-g: It could be useful to define an enum private to the Translation class purposed for just maintaining state aswell

lib/translator/translation.cc
harloff-g: This makes sense, I'll update accordingly

lib/translator/common.h
```diff
@@ -23,6 +28,34 @@ void DebugLog(const char* format, ...);
 
 void SetDebugCallback(void (*callback)(const char*));
 
+// Max consecutive pages required by NVMe PRP list is 512
+void* AllocPages(uint16_t count);
+
+void* DeallocPages(uint16_t count);
```
Ralith, line 34:
```suggestion
void* DeallocPages(void* pages, uint16_t count);
```
It's difficult to deallocate a resource without knowing which one it is.

lib/translator/common.h
```diff
@@ -23,6 +28,34 @@ void DebugLog(const char* format, ...);
...
+
+void SetPageCallback(void* (*alloc_callback)(uint16_t),
```
Ralith, line 36:
```suggestion
void SetPageCallbacks(void* (*alloc_callback)(uint16_t),
```

lib/translator/translation.cc
```diff
@@ -0,0 +1,69 @@
...
+    DebugLog("Null data_in buffer");
+    return StatusCode::kFailure;
+  }
+
+  scsi_defs::OpCode opc = static_cast<scsi_defs::OpCode>(this->scsi_cmd[0]);
+  switch (opc) {
+    case scsi_defs::OpCode::kInquiry:
+      return StatusCode::kSuccess;
+    default:
+      DebugLog("Bad OpCode");
```
Ralith, line 60: Let's print the hex value of the opcode here for ease of debugging. I expect that'll come in very useful for finding gaps when we test this live.

lib/translator/common.h
harloff-g: haha good point 😅

eyhhuang:
More specific implementations of keeping track of pages and deallocation should be separated out into a different PR if possible. For the translation library, it just needs to have an interface for requesting a page.

harloff-g:
> More specific implementations of keeping track of pages and deallocation should be separated out into a different PR if possible. For the translation library, it just needs to have an interface for requesting a page.

I agree with this. Is there anything currently in this PR that you feel is too specific or is it general enough?

eyhhuang:
> > More specific implementations of keeping track of pages and deallocation should be separated out into a different PR if possible. For the translation library, it just needs to have an interface for requesting a page.
> 
> I agree with this. Is there anything currently in this PR that you feel is too specific or is it general enough?

Nope! This PR's scope is good as is

lib/translator/common.h
```diff
@@ -15,14 +15,61 @@
...
+};
 
 void DebugLog(const char* format, ...);
 
 void SetDebugCallback(void (*callback)(const char*));
 
+// Max consecutive pages required by NVMe PRP list is 512
+void* AllocPages(uint16_t count);
+
+void* DeallocPages(void* pages_ptr, uint16_t count);
```
Ralith, line 42:
```suggestion
void DeallocPages(void* pages_ptr, uint16_t count);
```
I don't think there's a use for a return value here (or in the underlying callback).

lib/translator/translation.cc
```diff
@@ -0,0 +1,82 @@
...
+
+#include "common.h"
+
+namespace translator {
+
+BeginResponse Translation::Begin(absl::Span<const uint8_t> scsi_cmd,
+                                 scsi_defs::LunAddress lun) {
+  BeginResponse response = {};
+  response.status = ApiStatus::kSuccess;
+  if (this->pipeline_status != StatusCode::kUninitialized) {
```
Ralith, line 23:
```suggestion
  if (pipeline_status != StatusCode::kUninitialized) {
```
Member variables are in scope; they don't need to be qualified unless they're shadowed by something. Similar issues below.

lib/translator/common.h
```diff
@@ -15,14 +15,61 @@
...
+  // Returns a span containing translated NVMe commands.
+  absl::Span<const nvme_defs::GenericQueueEntryCmd> GetNvmeCmds();
+  // Aborts a given pipeline sequence and cleans up memory
+  void AbortPipeline();
+
+ private:
+  StatusCode pipeline_status = StatusCode::kUninitialized;
+  absl::Span<const uint8_t> scsi_cmd;
+  uint32_t nvme_cmd_count;
+  nvme_defs::GenericQueueEntryCmd nvme_cmds[3];
```
Ralith, line 70: Member variables should be named with a trailing underscore (see [Variable Names](https://google.github.io/styleguide/cppguide.html#Variable_Names) in the style guide).

lib/translator/common.h
```diff
@@ -15,14 +15,61 @@
...
 void DebugLog(const char* format, ...);
 
 void SetDebugCallback(void (*callback)(const char*));
 
+// Max consecutive pages required by NVMe PRP list is 512
+void* AllocPages(uint16_t count);
+
+void DeallocPages(void* pages_ptr, uint16_t count);
+
+void SetPageCallbacks(void* (*alloc_callback)(uint16_t),
```
Ralith, line 44: The definition uses the name `SetAllocPageCallbacks` but this prototype uses the name `SetPageCallbacks`. This mismatch will produce linker errors.

lib/translator/common.h
```diff
@@ -19,15 +19,31 @@
 #include <type_traits>
 
 #include "absl/types/span.h"
+#include "lib/scsi_defs.h"
+#include "third_party/spdk_defs/nvme_defs.h"
 
 namespace translator {
 
-enum class StatusCode { kSuccess, kInvalidInput, kNoTranslation, kFailure };
+enum class StatusCode {
```
eyhhuang, line 27: nit: might be helpful for future readability to add a comment about the purpose of StatusCode and similarly with ApiStatus. e.g. StatusCode is internal to the translation library

lib/translator/common.h
```diff
@@ -19,15 +19,31 @@
 #include <type_traits>
 
 #include "absl/types/span.h"
+#include "lib/scsi_defs.h"
+#include "third_party/spdk_defs/nvme_defs.h"
```
eyhhuang, line 23: nit: now that clang-format has `IncludeBlocks: preserve` we should probably have separate groups for other project's .h files, including third_party, and our own project's .h files. 

Although the style guide doesn't indicate a blank line between those groups, it later also says to have a blank line between each non-empty group. And in an earlier PR, I received a comment to separate them.

lib/translator/common.cc
```diff
@@ -35,4 +37,20 @@ void SetDebugCallback(void (*callback)(const char*)) {
   debug_callback = callback;
 }
```
eyhhuang, line 39: Can we have an abstraction for these three lines here https://github.com/googleinterns/scsi2nvme/pull/72/files#diff-b58a8575f98a080928fdb2d78d86ca29R64-R66 . Maybe `GetPrp()` but it's also used for metadata pointer so maybe a more generic name

lib/translator/translation.cc
```diff
@@ -0,0 +1,86 @@
...
+  if (scsi_cmd.empty()) {
+    DebugLog("Empty SCSI Buffer");
+    pipeline_status_ = StatusCode::kFailure;
+    return response;
+  }
+
+  pipeline_status_ = StatusCode::kSuccess;
+  scsi_cmd_ = scsi_cmd;
+
+  scsi_defs::OpCode opc = static_cast<scsi_defs::OpCode>(scsi_cmd[0]);
```
eyhhuang, line 39: Wondering if it's worth having a `raw_scsi = scsi_cmd_.subspan(1)` or something so we don't have to write `.subspan(1)` in every case.

test/translator/translation_test.cc
```diff
@@ -0,0 +1,56 @@
...
+
+TEST(Translation, ShouldHandleUnknownOpcode) {
+  translator::Translation translation = {};
+  uint8_t opc = 233;
+  absl::Span<const uint8_t> scsi_cmd = absl::MakeSpan(&opc, 1);
+  translator::BeginResponse resp = translation.Begin(scsi_cmd, 0);
+  EXPECT_EQ(translator::ApiStatus::kSuccess, resp.status);
+}
+
+TEST(Translation, ShouldReturnInquirySuccess) {
```
eyhhuang, line 33: this test will need to be updated once `InquiryToNvme` is implemented right? Because having a `scsi_cmd` of size 1 will will `InquiryToNvme` to return a `kInvalidInput`.

As a followup, when I'm implementing a translation function like `ReadToNvme`, should I add a case for that test here

harloff-g: Yeah this test was a POC to validate the Begin function's logical flow, it will have to be removed/replaced in InquiryToNvme. Imo I don't think it's necessary to have a test case for every single switch statement, assuming that the individual switch translation functions are well tested

lib/translator/translation.cc
harloff-g: I'll add this in, it seems like it will most likely make everyone's life easier when testing

lib/translator/common.cc
harloff-g: How about returning reinterpret_cast<uint64_t>(alloc_pages_callback(count)); ? I can't identify any cases where we actually want a void* response. An integer value of 0 implies the pointer was null.

eyhhuang: yeah that works.

lib/translator/common.cc
```diff
@@ -35,6 +37,25 @@ void SetDebugCallback(void (*callback)(const char*)) {
...
 
+// Returns void* casted to uint64_t.
+// A return value of 0 is equivalent to nullptr.
+uint64_t AllocPages(uint16_t count) {
```
Ralith, line 42: I'm guessing this is `uint64_t` because that's what NVMe wants, and we don't need to actually touch memory ourselves? I like that change, to the extent that I'd go ahead and use the same convention in the callbacks, then, unless there's something I'm missing.

lib/translator/translation.h
```diff
@@ -0,0 +1,55 @@
...
+#include "common.h"
+
+namespace translator {
+
+// Reports if the user is using the API correctly.
+enum class ApiStatus { kSuccess, kFailure };
+
+struct BeginResponse {
+  ApiStatus status;
+  uint32_t alloc_len;
```
Ralith, line 29: I'm guessing this is intended to define the length of `buffer` passed to `Complete`? A comment should document that.

lib/translator/common.h
```diff
@@ -20,16 +20,33 @@
...
 #include "absl/types/span.h"
+
 #include "lib/scsi_defs.h"
+#include "third_party/spdk_defs/nvme_defs.h"
 
 namespace translator {
 
-enum class StatusCode { kSuccess, kInvalidInput, kNoTranslation, kFailure };
+// Reports the status of a translation for internal use
+enum class StatusCode {
```
Ralith, line 30: nit: Maybe rename this to `TranslationState`? Also consider making it a private member type inside `Translation`, if it's no longer in the public API at all.

Ralith: Nevermind, this is used by downstream translation code.

PR #88, opened 2020-06-30
Translate Inquiry Command - Device Identification VPD

basimsahaf:
This PR contains code that returns the DeviceIdentification Page to the application client

TODO:
- [x] Merge with #166 to get the new WriteValue function


test/translator/inquiry_test.cc
```diff
@@ -235,11 +235,54 @@ TEST(TranslateUnitSerialNumberVpd, both) {
...
+      scsi_defs::PeripheralDeviceType::kDirectAccessBlock);
+
+  ASSERT_EQ(result.page_code, scsi_defs::PageCode::kDeviceIdentification);
+}
+
+TEST(TranslateIdentificationDescriptorNGUID, ReturnsCorrectFields) {
+  nvme_defs::IdentifyNamespace identify_namespace_data;
+  identify_namespace_data.eui64 = 0x0;
+  identify_namespace_data.nguid[0] = 0x123456789abcdefa;
+  identify_namespace_data.nguid[1] = 0x123456789abcdefa;
```
JustinAWei, line 263: are we actually checking this value in the result?

test/translator/inquiry_test.cc
```diff
@@ -235,11 +235,54 @@ TEST(TranslateUnitSerialNumberVpd, both) {
...
+  ASSERT_EQ(result.code_set, 0x1);
+  ASSERT_EQ(result.piv, 0);
+  ASSERT_EQ(result.association, 0b00);
+  ASSERT_EQ(result.identifier_type, 0x2);
+  ASSERT_EQ(result.identifier_length, 0x10);
+}
+
+TEST(TranslateIdentificationDescriptorEUI64, ReturnsCorrectFields) {
+  nvme_defs::IdentifyNamespace identify_namespace_data;
+  identify_namespace_data.eui64 = 0x123456789abcdefa;
```
JustinAWei, line 276: are we actually checking this value in the result?

lib/translator/inquiry.cc
```diff
@@ -236,6 +232,70 @@ scsi_defs::UnitSerialNumber BuildUnitSerialNumberVpd() {
...
+// spec: https://nvmexpress.org/wp-content/uploads/NVM_Express_-_SCSI_Translation_Reference-1_5_20150624_Gold.pdf
+// This only builds the constant part of the Identification Descriptor. The 
+// caller is responsible for building the identifier.
+scsi_defs::IdentificationDescriptor BuildIdentificationDescriptor(
+  nvme_defs::IdentifyNamespace &identify_namespace_data) {
+  scsi_defs::IdentificationDescriptor identification_descriptor = 
+    scsi_defs::IdentificationDescriptor();
+  identification_descriptor.protocol_identifier = 0x0;
+  identification_descriptor.code_set = 0x1;
+  identification_descriptor.piv = 0;
```
JustinAWei, line 245: use designated initialization
`struct point p = { .y = yvalue, .x = xvalue };`

lib/translator/inquiry.cc
```diff
@@ -236,6 +232,70 @@ scsi_defs::UnitSerialNumber BuildUnitSerialNumberVpd() {
...
+  // check if nonzero
+  bool nguid_nz =
+      identify_namespace_data.nguid[0] || identify_namespace_data.nguid[1];
+  bool eui64_nz = identify_namespace_data.eui64;
+
+  if (nguid_nz) {
+    identification_descriptor.identifier_length = 0x10;
+    uint8_t identification_descriptor_identifier[16];
+    memcpy(&identification_descriptor_identifier, &identify_namespace_data.nguid,
+      16);
```
JustinAWei, line 262: sizeof

lib/translator/inquiry.cc
```diff
@@ -236,6 +232,70 @@ scsi_defs::UnitSerialNumber BuildUnitSerialNumberVpd() {
...
+  // after we figure out how to send data back to SCSI. EUI-64 field is a 
+  // variable field hence its left out for now
+  
+  // check if nonzero
+  bool nguid_nz =
+      identify_namespace_data.nguid[0] || identify_namespace_data.nguid[1];
+  bool eui64_nz = identify_namespace_data.eui64;
+
+  if (nguid_nz) {
+    identification_descriptor.identifier_length = 0x10;
```
JustinAWei, line 259: make a constant for this `k..`

lib/translator/inquiry.cc
```diff
@@ -236,6 +232,70 @@ scsi_defs::UnitSerialNumber BuildUnitSerialNumberVpd() {
...
+    uint8_t identification_descriptor_identifier[16];
+    memcpy(&identification_descriptor_identifier, &identify_namespace_data.nguid,
+      16);
+  } else if(eui64_nz) {
+    identification_descriptor.identifier_length = 0x8;
```
JustinAWei, line 264: make constant for len

JustinAWei: or sizeof ui64

lib/translator/inquiry.cc
JustinAWei: or use sizeof

lib/translator/inquiry.cc
```diff
@@ -236,6 +232,70 @@ scsi_defs::UnitSerialNumber BuildUnitSerialNumberVpd() {
...
+  }
+  return identification_descriptor;
+}
+
+
+// Refer to section 6.1.2 on https://nvmexpress.org/wp-content/uploads/NVM_Express_-_SCSI_Translation_Reference-1_5_20150624_Gold.pdf
+// TODO:basimsahaf - return variable field at the end of the buffer.
+scsi_defs::DeviceIdentificationVpd BuildDeviceIdentificationVpd() {
+  scsi_defs::DeviceIdentificationVpd result = 
+    scsi_defs::DeviceIdentificationVpd();
```
JustinAWei, line 279: use designated init here

JustinAWei:
I added support for writing the SCSI result into a buffer. merge with `inquiry-standard` and check it out

lib/translator/inquiry.cc
basimsahaf: looks like I can't use `sizeof` for a bitfield. Do you suggest making an enum for it or a constant?

harloff-g: If the constant is just for inquiry.cc you could put it in the anonymous namespace so it's not exposed to the rest of the library

test/translator/inquiry_test.cc
basimsahaf: not the exact value but I'm checking whether `eui64` and `nguid` is zero or not

test/translator/inquiry_test.cc
basimsahaf: not the exact value but I'm checking whether `eui64` and `nguid` is zero or not

lib/translator/inquiry.cc
basimsahaf: updated!

lib/translator/inquiry.cc
basimsahaf: updated!

lib/translator/inquiry.cc
basimsahaf: updated!

lib/translator/inquiry.cc
basimsahaf: sounds good! I put the constant inside the anon namespace

basimsahaf:
> I added support for writing the SCSI result into a buffer. merge with `inquiry-standard` and check it out

Updated the code to support the buffer

lib/translator/inquiry.cc
```diff
@@ -13,10 +13,90 @@
...
+      /*
+      Shall be set to 00b indicating DESIGNATOR field is associated with logical
+      unit.
+      */
+      .association = 0b00,
+
+      /*
+      Shall be set to 2h indicating EUI-64 based identifier.
+      */
+      .identifier_type = 0x2,
```
oicirtap, line 93: It is good practice to have named variables for constants in your code (even if they are used only one or a few times). This makes code easier to read, and ensures that multiple uses of the same constant are guaranteed to be consistent (checking for these values in your tests for example).

Constants like `0b00 `, or `0x2 ` don't mean anything to the average reader, so having descriptive names for them make the piece of code much easier to understand.

test/translator/inquiry_test.cc
```diff
@@ -284,6 +285,76 @@ TEST(TranslateUnitSerialNumberVpd, both) {
   ASSERT_STREQ((char *)product_serial_number, formatted_hex_string);
 }
 
+TEST(TranslateIdentificationDescriptorNGUID, ReturnsCorrectFields) {
```
oicirtap, line 288: Perhaps a better test naming convention could be to name the test suite name (first argument) after the method being tested, and the second argument (test name) describing the specifics about what the test is testing. `ReturnsCorrectFields ` is not very descriptive as a test name.

lib/translator/inquiry.cc
```diff
@@ -13,10 +13,90 @@
...
 #include "inquiry.h"
+
 #include "string.h"
 
 namespace translator {
 
+// anonymous namespace for helper functions and constant expressions
+namespace {
+constexpr uint8_t identifier_length_eui64 = 0x8;
+constexpr uint8_t identifier_length_nguid = 0x10;
```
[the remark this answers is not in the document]
basimsahaf: Thanks for catching that. I changed the style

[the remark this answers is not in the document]
basimsahaf: That's a great suggestion. I moved them in the header file and now the test file also uses the constants

lib/translator/inquiry.cc
basimsahaf: That's a great suggestion. I realized that we can have enums for all these fields and therefore I added them. Let me know what you think about that

test/translator/inquiry_test.cc
basimsahaf: Yup, I agree. I made the test names more descriptive

lib/translator/inquiry.cc
```diff
@@ -13,10 +13,104 @@
...
+  if (raw_cmd.empty()) {
+    DebugLog("buffer is empty or nullptr");
+    return StatusCode::kInvalidInput;
+  }
+
+  uint8_t raw_opcode = raw_cmd[0];
+  if (raw_opcode > 0xaf) {
+    DebugLog("raw opcode is out of range: %x", raw_opcode);
+    return StatusCode::kInvalidInput;
+  }
```
eyhhuang, line 97: I'm not sure if this is just old code as a result of not merging in the base branch changes, but these checks (including whether the opcode == kInquiry) is no longer needed because it is covered by  Translation::Begin

lib/translator/inquiry.cc
```diff
@@ -13,10 +13,104 @@
...
+  if (opcode != scsi_defs::OpCode::kInquiry) {
+    absl::string_view expected_cmd_str =
+        ScsiOpcodeToString(scsi_defs::OpCode::kInquiry);
+    absl::string_view cmd_str = ScsiOpcodeToString(opcode);
+    DebugLog("wrong opcode. expected %s got %s.", expected_cmd_str, cmd_str);
+    return StatusCode::kInvalidInput;
+  }
+
+  // TODO: check invalid parameters in scsi_command
+  cmd = *reinterpret_cast<const scsi_defs::InquiryCommand *>(&raw_cmd[1]);
```
eyhhuang, line 110: Should use ReadValue https://github.com/googleinterns/scsi2nvme/blob/master/lib/translator/common.h#L32

lib/translator/inquiry.cc
```diff
@@ -158,6 +252,40 @@ void TranslateUnitSerialNumberVpd(
...
+      // Shall be set to 83h indicating Device Identification VPD Page, refer
+      // to 6.1.4
+      .page_code = scsi_defs::PageCode::kDeviceIdentification,
+
+      // Shall be set to the size of the remaining bytes of Device
+      // Identification
+      // VPD Page.
+      .page_length = sizeof(identification_descriptor),
+  };
+  memcpy(buffer.data(), &result, sizeof(result));
```
eyhhuang, line 284: Should use WriteValue

lib/translator/inquiry.h
```diff
@@ -25,6 +25,12 @@
 
 namespace translator {
 
+// constants for IdentifierLength field of DeviceIdentificationVPD
+constexpr uint8_t kIdentifierLengthEui64 = 0x8;
+constexpr uint8_t kIdentifierLengthNguid = 0x10;
+
+StatusCode RawToScsiCommand(absl::Span<const uint8_t>,
```
eyhhuang, line 32: nit: other places have been including variable names in the .h declarations. Although they're not necessary, as @JustinAWei  said, they could provide helpful context esp when the type is a `absl::Span<const uint8_t>`

lib/scsi_defs.h
```diff
@@ -195,6 +195,100 @@ enum class PageCode : uint8_t {
   kLogicalBlockProvisioningVpd = 0xb2
 };
 
+// SCSI Reference Manual Table 461
+// https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+enum class Association : uint8_t {
```
JustinAWei, line 200: great structs but seems extra in the future tbh unless we have code reuse. seems to me that inquiry structs returned are one-off

lib/translator/inquiry.cc
JustinAWei: it is unused, @basimsahaf check out `standard-inquiry` for updated inquiry code integrated w the generic translator

lib/scsi_defs.h
basimsahaf: I agree that it could be one off but it increases the readability of the code quite a lot. I made this change after @oicirtap suggested that I should make variables for all the struct field initializers. Although, other people might feel different.

CC: @eyhhuang @eyhhuang @Ralith @toadchild @w-michal

toadchild: I'm generally in favor of symbolic constants over magic numbers in code, even if they aren't used repeatedly.

lib/translator/inquiry.cc
```diff
@@ -13,12 +13,75 @@
 // limitations under the License.
 
 #include "inquiry.h"
+
 #include "string.h"
```
harloff-g, line 17: nit: #include <cstring> is more c++

lib/translator/inquiry.cc
```diff
@@ -13,12 +13,75 @@
...
+    memcpy(&identification_descriptor_identifier,
+           &identify_namespace_data.nguid,
+           sizeof(identification_descriptor_identifier));
+
+  } else if (eui64_nz) {
+    // Shall be set to 10h if the EUI-64 field is populated from the NVMe NGUID
+    // field.
+    identifier_length = kIdentifierLengthEui64;
+
+    uint64_t identification_descriptor_identifier =
```
harloff-g, line 53: What is the purpose of declaring this local variable and then immediately having it go out of scope?

lib/translator/inquiry.cc
```diff
@@ -13,12 +13,75 @@
...
+  uint8_t identifier_length;
+
+  if (nguid_nz) {
+    // Shall be set to 08h if the EUI-64 field is populated from the NVMe EUI64
+    // field.
+    identifier_length = kIdentifierLengthNguid;
+
+    uint8_t identification_descriptor_identifier[16];
+
+    memcpy(&identification_descriptor_identifier,
```
harloff-g, line 44: Why do we memcpy this data into identification_descriptor_identifier and do nothing with it

basimsahaf: This function returns the identificaiton descriptor to the `TranslateDeviceIdentification` [function](https://github.com/googleinterns/scsi2nvme/pull/128/commits/5fc032b0e0019378dcf8d24be631e780c13b5480) which then [writes](https://github.com/googleinterns/scsi2nvme/pull/128/commits/5fc032b0e0019378dcf8d24be631e780c13b5480) it to the `buffer`.

lib/scsi_defs.h
```diff
@@ -874,6 +978,56 @@ struct UnmapBlockDescriptor {
...
+  uint8_t _reserved : 8;
+  uint8_t page_length : 8;
+  // uint8_t product_serial_number[256];
+} ABSL_ATTRIBUTE_PACKED;
+static_assert(sizeof(UnitSerialNumber) == 4);
+
+// SCSI Reference Manual Table 460
+// https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+struct IdentificationDescriptor {
+  ProtocolIdentifier protocol_identifier : 4;
```
toadchild, line 1008: I think you have some fields out of order - the code set is bits 0-4 and the protocol identifier is bits 5-7.

basimsahaf: thanks for pointing that out. Fixed it

lib/translator/inquiry.cc
basimsahaf: Yup, I was doing something that didn't work. I've updated the code now. Let me know your thoughts

lib/translator/inquiry.cc
basimsahaf: https://github.com/googleinterns/scsi2nvme/pull/88#discussion_r461218680

lib/translator/inquiry.cc
```diff
@@ -25,6 +25,73 @@ namespace translator {
...
+  // Writes the value of the identification at the end of the buffer. This data
+  // comes after the DeviceIdentificationVPD and IdentificationDescriptor
+  // struct. Hence, the data is written at index +
+  // sizeof(IdentificationDescriptor), where index is the size of the
+  // DeviceIdentificationVPD struct.
+  if (nguid_nz) {
+    // Shall be set to 08h if the EUI-64 field is populated from the NVMe EUI64
+    // field.
+    identifier_length = kIdentifierLengthNguid;
+    WriteValue(identify_namespace_data.nguid,
```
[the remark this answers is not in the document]
basimsahaf: That's a good idea! Adding that in

lib/translator/inquiry.cc
```diff
@@ -25,6 +25,83 @@ namespace translator {
...
+constexpr uint8_t kIdentifierLengthEui64 = 0x8;
+
+// Builds the IdentificationDescriptor according to 6.1.4.5 of the translation
+// spec:
+// https://nvmexpress.org/wp-content/uploads/NVM_Express_-_SCSI_Translation_Reference-1_5_20150624_Gold.pdf
+// This only builds the constant part of the Identification Descriptor. The
+// caller is responsible for building the identifier.
+StatusCode BuildIdentificationDescriptor(
+    const nvme::IdentifyNamespace& identify_namespace_data,
+    Span<uint8_t> buffer, int index, uint8_t& identifier_length) {
```
Ralith, line 39: This function seems to only write to `buffer` starting at `index`. Would it be simpler for the caller to pass in `buffer.subspan(index)` from the start?

lib/translator/inquiry.cc
```diff
@@ -158,6 +235,48 @@ void TranslateUnitSerialNumberVpd(
   WriteValue(product_serial_number, buffer.subspan(sizeof(result)));
 }
 
+// Refer to section 6.1.2 on
+// https://nvmexpress.org/wp-content/uploads/NVM_Express_-_SCSI_Translation_Reference-1_5_20150624_Gold.pdf
+StatusCode TranslateDeviceIdentificationVpd(
+    const nvme::IdentifyNamespace& identify_namespace, Span<uint8_t> buffer) {
+  uint8_t identification_descirptor_length = 0;
```
Ralith, line 242:
```suggestion
  uint8_t identification_descriptor_length = 0;
```

lib/translator/inquiry.cc
basimsahaf: that's a great suggestion. Definitely cleaner. Updated the code

PR #89, opened 2020-06-30
Translate Inquiry Command - Block Limits VPD

JustinAWei:
Based on `inquiry-standard`. Merge `inquiry-standard` first. #76


lib/translator/inquiry.cc
```diff
@@ -236,6 +236,58 @@ scsi_defs::UnitSerialNumber BuildUnitSerialNumberVpd() {
   return TranslateUnitSerialNumberVpdResponse(identify_namespace_data);
 }
 
+scsi_defs::BlockLimitsVpd BuildBlockLimitsVpd() {
```
eyhhuang, line 239: Can this function be omitted and the caller just calls `TranslateBuildBlockLimitsVpdResponse` directly?

lib/translator/inquiry.cc
```diff
@@ -266,10 +318,11 @@ void translate(absl::Span<const uint8_t> raw_cmd) {
         // TODO: May optionally be supported by returning Extended INQUIRY data
         // page toapplication client, refer to 6.1.5.
         break;
-      case scsi_defs::PageCode::kBlockLimitsVpd:
-        // TODO: May be supported by returning Block Limits VPD data page to
+      case scsi_defs::PageCode::kBlockLimitsVpd: {
+        // May be supported by returning Block Limits VPD data page to
         // application client, refer to 6.1.6.
-        break;
+        scsi_defs::BlockLimitsVpd result = BuildBlockLimitsVpd();
```
eyhhuang, line 324: why is a variable declared in this scope?

JustinAWei: to be written to the buffer. handling the completed scsi response is not implemented yet. waiting on public interface.

lib/translator/inquiry.cc
```diff
@@ -187,6 +187,56 @@ void TranslateUnitSerialNumberVpd(
   memcpy(&buffer[sizeof(result)], &product_serial_number, result.page_length);
 }
 
+void TranslateBuildBlockLimitsVpd(
```
eyhhuang, line 190: I'm not too sure how the inquiry command works but will this function be part of the interface called by the engine or is it mainly a helper function? If it is part of the interface, can we add a `ToScsi` suffix to the name?

Of if you're against the `ToScsi`, I feel like having both `Translate` and `Build` in the name is confusing so maybe just have one

lib/translator/inquiry.cc
```diff
@@ -187,6 +187,56 @@ void TranslateUnitSerialNumberVpd(
...
+    const nvme_defs::IdentifyControllerData &identify_controller_data,
+    absl::Span<uint8_t> buffer) {
+  uint32_t max_transfer_length =
+      identify_controller_data.mdts ? 1 << identify_controller_data.mdts : 0;
```
eyhhuang, line 194: can we add a comment somewhere that the mdts represents the size as 2^n. I was a little confused by this bit shift initially. Also mdts is at most 2^8, while the size of max_transfer_length is 4 bytes. Is there a concern of overflow?

lib/translator/inquiry.cc
```diff
@@ -220,10 +270,11 @@ void translate(absl::Span<const uint8_t> raw_cmd, absl::Span<uint8_t> buffer) {
         // TODO: May optionally be supported by returning Extended INQUIRY data
         // page toapplication client, refer to 6.1.5.
         break;
-      case scsi_defs::PageCode::kBlockLimitsVpd:
-        // TODO: May be supported by returning Block Limits VPD data page to
+      case scsi_defs::PageCode::kBlockLimitsVpd: {
```
eyhhuang, line 273: this case doesn't introduce a local variable so i dont think the { is needed

eyhhuang:
Loving the small PR size  🥳

lib/translator/inquiry.cc
JustinAWei: welll i hope nobody expects a page length of 2^2^8
good point though i'll look into it

JustinAWei: I'll add a bounds check

lib/translator/inquiry.cc
```diff
@@ -187,6 +187,64 @@ void TranslateUnitSerialNumberVpd(
...
+  */
+  if (identify_controller_data.mdts > 64) {
+    DebugLog("max transfer length is > 2^64");
+  }
+  uint32_t max_transfer_length =
+      identify_controller_data.mdts ? 1 << identify_controller_data.mdts : 0;
+
+  scsi_defs::BlockLimitsVpd result = {
+      .page_code = scsi_defs::PageCode::kBlockLimitsVpd,
+      .page_length = 0x003c,
```
oicirtap, line 206: Can we add a named variable for `0x003c `?

lib/translator/inquiry.cc
```diff
@@ -187,6 +187,64 @@ void TranslateUnitSerialNumberVpd(
...
+      .max_unmap_lba_count = identify_controller_data.oncs.dsm,
+
+      /*
+      Shall be set to 0000_0000h if Dataset Management
+      command – Deallocate (AD) attribute is not supported.
+      Shall be set to 0000_0100h if Dataset Management
+      command – Deallocate (AD) attribute is supported.
+      */
+      .max_unmap_block_descriptor_count =
+          identify_controller_data.oncs.dsm ? 0x0100 : 0};
```
oicirtap, line 243: Can we add a named variable for `0x0100 `?

test/translator/inquiry_test.cc
```diff
@@ -315,4 +315,27 @@ TEST(TranslateUnitSerialNumberVpd, neither) {
   ASSERT_STREQ(product_serial_number, formatted_hex_string);
 }
 
+TEST(TranslateBlockLimitsVpd, success) {
```
oicirtap, line 318: Can we add tests for the different possible contents of identify_controller_data?

lib/translator/inquiry.cc
JustinAWei: will make all the constants when we've consolidated the inquiry branches

lib/translator/inquiry.cc
```diff
@@ -151,6 +151,61 @@ void TranslateUnitSerialNumberVpd(
...
+
+void TranslateBlockLimitsVpd(const nvme::IdentifyControllerData& identify_ctrl,
+                             absl::Span<uint8_t> buffer) {
+  // The value is in units of the minimum memory
+  // page size (CAP.MPSMIN) and is reported as a power of two (2^n).
+  // A value of 0h indicates that there is no maximum data transfer size
+  uint32_t max_transfer_length;
+  if (identify_ctrl.mdts > 32) {
+    DebugLog("max transfer length is > 2^32");
+    max_transfer_length = 1 << 31;
```
toadchild, line 163: There's no need to preserve the power-of-two nature here and you could return the max value for a uint32 (2^32 - 1)

lib/translator/inquiry.cc
```diff
@@ -151,6 +151,61 @@ void TranslateUnitSerialNumberVpd(
...
+      // to the value in MAXIMUM TRANSFER LENGTH field if
+      // Fused Operation is supported.
+      .max_compare_write_length =
+          identify_ctrl.fuses.compare_and_write ? compare_and_write_len : 0,
+
+      // Shall be set to value calculated according to method
+      // described in NVMe v1.1 Identify Controller Data
+      // Structure: Maximum Data Transfer Size (MDTS)
+
+      // The maximum transfer data size is reported as 2^n
```
toadchild, line 190: This field is a number of blocks, not an exponent value.

lib/translator/inquiry.cc
harloff-g: Max transfer length should be 2^16 according to section 3.13 of https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf

lib/translator/inquiry.cc
harloff-g: Best practice to add a TODO in that case

lib/translator/inquiry.cc
```diff
@@ -152,6 +152,66 @@ void TranslateUnitSerialNumberVpd(
...
+                             absl::Span<uint8_t> buffer) {
+  // The value is in units of the minimum memory
+  // page size (CAP.MPSMIN) and is reported as a power of two (2^n).
+  // A value of 0h indicates that there is no maximum data transfer size
+  uint32_t max_transfer_length;
+  if (identify_ctrl.mdts > 32) {
+    // Max transfer length should be 2^16 according to section 3.13 of
+    // https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+    DebugLog("max transfer length is > 2^32");
+    max_transfer_length = 1 << 16;
```
oicirtap, line 166: If power of 2's will be frequent (ish), we could add a macro or helper function to perform them, making the code more readable. Just an observation.

lib/translator/inquiry.cc
```diff
@@ -152,6 +152,66 @@ void TranslateUnitSerialNumberVpd(
...
+  } else {
+    max_transfer_length = identify_ctrl.mdts ? 1 << identify_ctrl.mdts : 0;
+  }
+
+  // TODO: put in common?
+  const uint8_t kMaxCompareWriteLen = 255;
+  // compare and write length is 8 bits. we ensure the size fits.
+  uint32_t compare_and_write_len = (max_transfer_length > kMaxCompareWriteLen)
+                                       ? kMaxCompareWriteLen
+                                       : max_transfer_length;
```
oicirtap, line 176: Similarly, we could define max and min functions if you think they'd improve readability. We could add a number of utility functions in a different PR. Or there might already exists some bazel target with common definitions we can use.

lib/translator/inquiry.cc
JustinAWei: will look into doing a macro on another pr

lib/translator/inquiry.cc
JustinAWei: interesting, will create utilis issue

PR #91, opened 2020-06-30
Translate Inquiry Command - Logical Block Provisioning VPD Page

JustinAWei:
Based on `inquiry-standard`. Merge `inquiry-standard` first.


lib/translator/inquiry.cc
```diff
@@ -228,7 +331,10 @@ void translate(absl::Span<const uint8_t> raw_cmd, absl::Span<uint8_t> buffer) {
         // TODO: Return Block Device Characteristics Vpd Page to application
         // client, refer to 6.1.7.
         break;
-      case scsi_defs::PageCode::kLogicalBlockProvisioningVpd:
+      case scsi_defs::PageCode::kLogicalBlockProvisioningVpd: {
```
eyhhuang, line 334: nit: no need to { here

lib/translator/inquiry.cc
```diff
@@ -187,6 +187,109 @@ void TranslateUnitSerialNumberVpd(
...
+  command – Deallocate (AD) attribute is supported and
+  Identify Namespace Data – NSFEAT bit 0 is reported “0”,,
+  indicating that the namespace is resource-provisioned.
+
+  Shall be set to 2(Thin) if Dataset Management command
+  – Deallocate (AD) attribute is supported and Identify
+  Namespace Data – NSFEAT bit 0 is reported “1”,
+  indicating that the namespace is thin-provisioned.
+   */
+  if (!ad && !identify_namespace_data.nsfeat.thin_prov)
```
toadchild, line 266: I have a strong personal preference against C/C++ code using if/else with no braces, but I don't know that it officially violates the style rules you're using for this project.

lib/translator/inquiry.cc
```diff
@@ -187,6 +187,109 @@ void TranslateUnitSerialNumberVpd(
...
 
+void TranslateLogicalBlockProvisioningVpd(
+    const nvme_defs::IdentifyControllerData &identify_controller_data,
+    const nvme_defs::IdentifyNamespace &identify_namespace_data,
+    absl::Span<uint8_t> buffer) {
+  bool ad = identify_controller_data.oncs.dsm;
+
+  scsi_defs::LogicalBlockProvisioningVpd result = {
+      .page_code = scsi_defs::PageCode::kBlockLimitsVpd,
+      .page_length = 0x04,
```
oicirtap, line 198: Can we add a named variable for `0x04 `?

Since there are so many possible values for `page_length ` throughout the different components of the inquiry command, we can also consider creating an enum for them in the header. We could even set the type of `page_length` to said enum for type safety. Up to you though.

lib/translator/inquiry.cc
```diff
@@ -187,6 +187,109 @@ void TranslateUnitSerialNumberVpd(
...
+      TODO: Lack of support from thin provisioning by itself may not
+      guarantee lack of resource provisioning. There may be an explicit flag
+      about resource provisioning.
+      */
+
+      .anc_sup = identify_namespace_data.nsfeat.thin_prov,
+
+  };
+
+  /*
```
oicirtap, line 249: You tend to space your comment lines on the shorter side. While this is not per se an issue, it does lead to longer than needed comment blocks and longer files.

The maximum line length according to the style guide is 80 characters, so feel free to write a bit longer comment lines if you want.

lib/translator/inquiry.cc
oicirtap: I would either add braces, or remove the blank lines between clauses. Blank lines in code tend to create a mental disconnect between the stuff above and below, but in if/else clauses it should be clear which clauses are linked to each other.

test/translator/inquiry_test.cc
```diff
@@ -284,6 +284,39 @@ TEST(TranslateUnitSerialNumberVpd, both) {
   ASSERT_STREQ((char *)product_serial_number, formatted_hex_string);
 }
 
+TEST(TranslateLogicalBlockProvisioningVpd, success) {
```
oicirtap, line 287: Can we add different tests for different codepaths?

lib/translator/inquiry.cc
```diff
@@ -187,6 +187,109 @@ void TranslateUnitSerialNumberVpd(
   memcpy(&buffer[sizeof(result)], &product_serial_number, result.page_length);
 }
 
+void TranslateLogicalBlockProvisioningVpd(
+    const nvme_defs::IdentifyControllerData &identify_controller_data,
+    const nvme_defs::IdentifyNamespace &identify_namespace_data,
+    absl::Span<uint8_t> buffer) {
```
basimsahaf, line 193: Shouldn't buffer be sent by reference here since we are writing to the buffer we create in the `translate` function below.?

JustinAWei: span is a wrapper around a pointer so even if its a copy, the span will still wrap around the same pointer

lib/translator/inquiry.cc
```diff
@@ -187,6 +187,109 @@ void TranslateUnitSerialNumberVpd(
...
+      LBPRZ
+      Shall be set to 1 if Dataset Management command –
+      Deallocate (AD) attribute is supported and the device
+      returns all zeros for reads of deallocated LBAs, otherwise
+      set to 0.
+
+      TODO: the AD attribute itself may not guarantee the device
+      returns all zeros for reads of deallocated LBAs.
+      May have to double check this one if there are any errors.
+      */
```
basimsahaf, line 226: nit: Google C++ style guide recommends sticking to one style of making comments on the file:

https://google.github.io/styleguide/cppguide.html#Comment_Style

lib/translator/inquiry.cc
JustinAWei: will add contstants once all inquiry branches consolidated

test/translator/inquiry_test.cc
```diff
@@ -330,4 +325,135 @@ TEST_F(InquiryTest, TranslateUnitSerialNumberVpdNone) {
...
+  }
+
+  if (ad || result.provisioning_type == 1 || result.provisioning_type == 2) {
+    EXPECT_EQ(result.lbpu, 1);
+  } else {
+    EXPECT_EQ(result.lbpu, 0);
+  }
+}
+
+}  // namespace
```
harloff-g, line 459: nit: newline at end of file

lib/translator/inquiry.cc
```diff
@@ -185,6 +185,88 @@ StatusCode InquiryToNvme(absl::Span<const uint8_t> raw_scsi,
...
+      // Shall be set to 1, to indicate that setting the ANCHOR bit in UNMAP is
+      // supported if the namespace is resource provisioned.
+
+      // May be set to 1 if the namespace is thin provisioned.
+
+      // TODO: Lack of support from thin provisioning by itself may not
+      // guarantee lack of resource provisioning. There may be an explicit flag
+      // about resource provisioning.
+
+      .anc_sup = identify_ns.nsfeat.thin_prov,
```
harloff-g, line 232: per @w-michal we are not currently supporting the anchor bit in unmap, so let's just always return 0 here.

test/translator/inquiry_test.cc
```diff
@@ -330,4 +325,135 @@ TEST_F(InquiryTest, TranslateUnitSerialNumberVpdNone) {
...
+  EXPECT_EQ(result.page_code, scsi::PageCode::kLogicalBlockProvisioningVpd);
+  EXPECT_EQ(result.lbprz, ad);
+  EXPECT_EQ(result.anc_sup, identify_ns.nsfeat.thin_prov);
+  if (!ad && !identify_ns.nsfeat.thin_prov) {
+    EXPECT_EQ(result.provisioning_type, 0);
+  } else if (ad && !identify_ns.nsfeat.thin_prov) {
+    EXPECT_EQ(result.provisioning_type, 1);
+  } else if (ad && identify_ns.nsfeat.thin_prov) {
+    EXPECT_EQ(result.provisioning_type, 2);
+  }
```
oicirtap, line 351: I'm not sure how I feel about having these if statements in your tests. Replicating the implementation logic in a test to test the implementation can be dangerous for two reasons:
1. if the implementation has a bug, the test will have the same bug and would fail to catch it.
2. Even if the specific replicated piece of logic is correct, when we test results conditionally like here, we might fail to notice if at some point the test starts taking path B instead of A (which might be undesirable).

Since your tests all define a specific `thin_prov ` and `ad` value combination, we can also have the test check only that specific path, instead of adding if clauses that will never be traversed by a given test. In this test for example it seems that both `add` and `thin_prov ` will always be `0`?

Issue #68, opened 2020-06-16
Translate Read Command


eyhhuang:
Translate SCSI Read 6, 10, 12, 16 command to NVMe Read command

Issue #93, opened 2020-07-01
Translate NVMe read response to SCSI


eyhhuang:
(title)

PR #96, opened 2020-07-01
Translate read

eyhhuang:
Closes https://github.com/googleinterns/scsi2nvme/issues/68


lib/translator/read.cc
```diff
@@ -0,0 +1,181 @@
...
+void LegacyRead(uint32_t nsid, uint32_t lba, uint16_t transfer_length,
+                nvme_defs::GenericQueueEntryCmd& queue_entry) {
+  queue_entry.opc = static_cast<uint8_t>(nvme_defs::NvmOpcode::kRead);
+  queue_entry.fuse = 0b00;  // Normal operation, not fused
+  queue_entry.psdt = 0b00;  // PRPs are used for data transfer
+  queue_entry.mptr = reinterpret_cast<std::uint64_t>(AllocPages(1));
+
+  queue_entry.nsid = nsid;
+
+  queue_entry.dptr.prp.prp1 = reinterpret_cast<std::uint64_t>(AllocPages(1));
```
harloff-g, line 71: Will need to define a way of constructing PRP lists as the read command is defined to use them in dptr field, per https://nvmexpress.org/wp-content/uploads/NVM-Express-1_4-2019.06.10-Ratified.pdf section 6.9

harloff-g: Given recent developments in communication with the NVMe device this may not be nessesary

lib/translator/read.cc
```diff
@@ -0,0 +1,181 @@
...
+}  // namespace
+
+StatusCode Read6ToNvme(uint32_t nsid, absl::Span<const uint8_t> raw_cmd,
+                       nvme_defs::GenericQueueEntryCmd queue_entry) {
+  if (raw_cmd.size() != sizeof(scsi_defs::Read6Command)) {
+    DebugLog("Malformed Read6 command");
+    return StatusCode::kInvalidInput;
+  }
+  scsi_defs::Read6Command cmd;
+  memcpy(&cmd, &raw_cmd, sizeof(scsi_defs::Read6Command));
```
Ralith, line 116: This is copying the representation of the `absl::Span`, not the data it points to. Use `raw_cmd.data()` to get the data it points to.

JustinAWei:
maybe diff this pr against translation instead of master to better see read translation specific code

lib/scsi_defs.h
```diff
@@ -389,7 +389,7 @@ static_assert(sizeof(Read6Command) == 5);
 // SCSI Reference Manual Table 97
 // https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
 struct Read10Command {
-  uint8_t rd_protect : 3;  // read protect bit
+  uint8_t rdprotect : 3;   // read protect bit
```
eyhhuang, line 392: renaming because the field is RDPROTECT so the underscore didn't make sense in the first place.

lib/scsi_defs.h
```diff
@@ -448,16 +448,16 @@ struct Read32Command {
...
   bool dpo : 1;     // disable page output bit
   bool fua : 1;     // Forced Unit access bit
   bool rarc : 1;    // Rebuild Assist Recovery bit
   bool fua_nv : 1;  // Forced Access Unit Non-Volatile
   uint16_t reserved_3 : 9;
   uint64_t logical_block_address : 64;
-  uint32_t expected_initial_logical_block_reference_tag : 32;
-  uint16_t expected_logical_block_application_tag : 16;
-  uint16_t logical_block_application_tag_mask : 16;
+  uint32_t eilbrt : 32;  // expected_initial_logical_block_reference_tag
```
eyhhuang, line 458: renaming for convenience. using the acronym is also consistent with how the docs reference this field

lib/scsi_defs.h
```diff
@@ -389,11 +389,11 @@ static_assert(sizeof(Read6Command) == 5);
 // SCSI Reference Manual Table 97
 // https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
 struct Read10Command {
-  uint8_t rd_protect : 3;  // read protect bit
-  bool dpo : 1;            // disable page output bit
-  bool fua : 1;            // Forced Unit access bit
-  bool rarc : 1;           // Rebuild Assist Recovery bit
-  bool fua_nv : 1;         // Forced Access Unit Non-Volatile
+  uint8_t rdprotect : 3;  // read protect bit
```
eyhhuang, line 392: renaming because the field is RDPROTECT so the underscore didn't make sense in the first place.

lib/scsi_defs.h
```diff
@@ -448,16 +448,16 @@ struct Read32Command {
...
   bool dpo : 1;     // disable page output bit
   bool fua : 1;     // Forced Unit access bit
   bool rarc : 1;    // Rebuild Assist Recovery bit
   bool fua_nv : 1;  // Forced Access Unit Non-Volatile
   uint16_t reserved_3 : 9;
   uint64_t logical_block_address : 64;
-  uint32_t expected_initial_logical_block_reference_tag : 32;
-  uint16_t expected_logical_block_application_tag : 16;
-  uint16_t logical_block_application_tag_mask : 16;
+  uint32_t eilbrt : 32;  // Expected Initial Logical Block Reference Tag
```
eyhhuang, line 458: renaming for convenience. using the acronym is also consistent with how the docs reference this field

.github/workflows/main.yml
```diff
@@ -34,6 +34,7 @@ jobs:
   
   build-and-test:
     env:
+      CC: gcc-8
```
eyhhuang, line 37: Change in master needed for this build to pass that hasn't been merged into the base branch yet.

lib/translator/read.cc
```diff
@@ -0,0 +1,203 @@
...
+}
+
+}  // namespace
+
+StatusCode Read6ToNvme(absl::Span<const uint8_t> scsi_cmd,
+                       nvme_defs::GenericQueueEntryCmd& nvme_cmd) {
+  if (scsi_cmd.size() != sizeof(scsi_defs::Read6Command)) {
+    DebugLog("Malformed Read6 command");
+    return StatusCode::kInvalidInput;
+  }
```
Ralith, line 115: This check, and the similar ones below, are redundant to the size check performed by `ReadValue`. We might as well just rely on that--though the debug messages associated with the first check in each case seem a bit clearer.

eyhhuang: Good point!

lib/translator/read.h
```diff
@@ -0,0 +1,57 @@
...
+#define LIB_TRANSLATOR_READ_H
+
+#include "third_party/spdk/nvme.h"
+
+#include "common.h"
+#include "lib/scsi.h"
+
+namespace translator {
+
+/**
```
harloff-g, line 25: nit: codebase currently uses // for comments rather than /*

lib/translator/read.cc
```diff
@@ -0,0 +1,178 @@
...
+
+StatusCode LegacyRead(uint32_t lba, uint16_t transfer_length,
+                      nvme::GenericQueueEntryCmd& nvme_cmd) {
+  nvme_cmd = nvme::GenericQueueEntryCmd{
+      .opc = static_cast<uint8_t>(nvme::NvmOpcode::kRead),
+      .psdt = 0b00  // PRPs are used for data transfer
+  };
+
+  uint64_t mptr = AllocPages(1);
+  uint64_t prp = AllocPages(1);
```
harloff-g, line 72: Add functionality to work with #105

lib/translator/read.cc
```diff
@@ -0,0 +1,183 @@
+#include "read.h"
+
+#include "absl/types/span.h"
+
+#include "common.h"
```
basimsahaf, line 5: nit: read.h and common.h includes can be put in two consecutive lines

lib/translator/read.cc
```diff
@@ -0,0 +1,183 @@
...
+
+  if (status != StatusCode::kSuccess) {
+    return status;
+  }
+
+  nvme_cmd.cdw[2] |= (prinfo << 26);  // cdw12 prinfo bits 29:26
+  nvme_cmd.cdw[2] |= (fua << 30);     // cdw12 fua bit 30;
+
+  return StatusCode::kSuccess;
+}
```
basimsahaf, line 106: I think these two functions need more documentation. It won't be immediately clear to the reader as to why Legacy Read exists.

eyhhuang: Good point! i'll definitely add some comments

lib/translator/read.cc
eyhhuang: I think the .h file that you're implementing is supposed to be in its own line at the top https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

lib/translator/read.cc
```diff
@@ -0,0 +1,187 @@
...
+namespace translator {
+
+namespace {  // anonymous namespace for helper functions
+
+// Section 5.3
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+StatusCode BuildPrinfo(uint8_t rdprotect, uint8_t& prinfo) {
+  bool pract;     // Protection Information Action 1 bit
+  uint8_t prchk;  // Protection Information Check 3 bits
+  prinfo = 0;     // Protection Information field 4 bits
```
toadchild, line 16: Small nit, but this will write to the out-parameter even if the following switch statement returns an error code.  In general, having out-parameters be undefined in the case of errors is ok, but it's something to consider.

lib/translator/read.cc
```diff
@@ -0,0 +1,187 @@
...
+
+  return StatusCode::kSuccess;
+}
+
+void SetLbaTags(uint32_t eilbrt, uint16_t elbat, uint16_t elbatm,
+                nvme::GenericQueueEntryCmd& nvme_cmd) {
+  // cdw14 expected initial logical block reference
+  nvme_cmd.cdw[4] = eilbrt;
+  // cdw15 bits 15:0 expected logical block application tag
+  nvme_cmd.cdw[5] |= elbat;
```
toadchild, line 59: Should this be assignment?  Are you intending to preserve bits that might already be set in cdw[5]?

lib/translator/read.cc
```diff
@@ -0,0 +1,187 @@
...
+                      Allocation& allocation) {
+  StatusCode status_code = allocation.SetPages(1, 1);
+
+  if (status_code != StatusCode::kSuccess) {
+    return status_code;
+  }
+
+  nvme_cmd = nvme::GenericQueueEntryCmd{
+      .opc = static_cast<uint8_t>(nvme::NvmOpcode::kRead),
+      .psdt = 0b00  // PRPs are used for data transfer
```
toadchild, line 78: Nit: any reason to not just use `0`?

lib/translator/read.cc
toadchild: Instead of setting this to zero and or-ing prchk, you could just assign prchk below.

lib/translator/read.cc
```diff
@@ -0,0 +1,187 @@
...
+  };
+
+  nvme_cmd.mptr = allocation.mdata_addr;
+  nvme_cmd.dptr.prp.prp1 = allocation.data_addr;
+  nvme_cmd.cdw[0] = lba;              // cdw10 Starting lba bits 31:00
+  nvme_cmd.cdw[2] = transfer_length;  // cdw12 nlb bits 15:00
```
toadchild, line 84: In nvme, the transfer length is a 0-based number of blocks; I.e. a value of 0 here means to read 1 block.  I think you'll need to include an adjustment factor here.

lib/translator/read.cc
```diff
@@ -0,0 +1,187 @@
...
+
+  nvme_cmd.cdw[2] |= (prinfo << 26);  // cdw12 prinfo bits 29:26
+  nvme_cmd.cdw[2] |= (fua << 30);     // cdw12 fua bit 30;
+
+  return StatusCode::kSuccess;
+}
+
+}  // namespace
+
+StatusCode Read6ToNvme(absl::Span<const uint8_t> scsi_cmd,
```
toadchild, line 114: It looks like Read6 has some special handling required for the transfer length - a length of 0 is treated as a length of 256.

lib/translator/read.cc
```diff
@@ -0,0 +1,187 @@
...
+  // cdw15 bits 15:0 expected logical block application tag
+  nvme_cmd.cdw[5] |= elbat;
+  // cdw15 bits 31:16 expected logical block application tag mask
+  nvme_cmd.cdw[5] |= (elbatm << 16);
+}
+
+// Translates fields common to all Read commands
+// Named Legacy because it is called directly by Read6, an obsolete command
+// lacking fields common to other Read commands
+StatusCode LegacyRead(uint32_t lba, uint16_t transfer_length,
```
toadchild, line 67: LBA needs to be 64 bit; Read16 and nvme both use 64 bit values.

toadchild: Read16 also has a 32 bit transfer length field.

lib/translator/read.cc
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  nvme_cmd = nvme::GenericQueueEntryCmd{
+      .opc = static_cast<uint8_t>(nvme::NvmOpcode::kRead),
+      .psdt = 0b00  // PRPs are used for data transfer
+  };
+
+  nvme_cmd.mptr = allocation.mdata_addr;
+  nvme_cmd.dptr.prp.prp1 = allocation.data_addr;
+  nvme_cmd.cdw[0] = lba;              // cdw10 Starting lba bits 31:00
```
toadchild, line 83: These fields may also need to be swapped between host and network byte order.

lib/translator/read.cc
eyhhuang: This should be assignment. I had it as an "or equals" so it would look consistent with line 61. But I definitely don't want to preserve any bits.

lib/translator/read.cc
eyhhuang: I thought using `0b00` would more clearly indicate that this field is 2 bits long but `0b00` clearly isn't as obvious an indicator as having a comment describing the field size. However, at this point it's probably not important to the reader the size of the field anyway, so i'll just go with '0'.

lib/translator/read.cc
eyhhuang: thanks for the catch!

lib/translator/read.cc
basimsahaf: oh true my bad. Missed that its read.cc

lib/translator/read.cc
```diff
@@ -0,0 +1,183 @@
...
+
+#include "common.h"
+
+namespace translator {
+
+namespace {  // anonymous namespace for helper functions
+
+// Section 5.3
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+StatusCode BuildPrinfo(uint8_t rdprotect, uint8_t& prinfo) {
```
basimsahaf, line 13: suggestion: we can move this function in common.h since write will also be using this

eyhhuang: Sounds good. I'll also open that change in a new PR

lib/translator/read.cc
eyhhuang: Endianess is definitely something I missed here! Going to discuss with the rest of the team how best to handle this

lib/translator/read.cc
```diff
@@ -0,0 +1,187 @@
...
+      .opc = static_cast<uint8_t>(nvme::NvmOpcode::kRead),
+      .psdt = 0b00  // PRPs are used for data transfer
+  };
+
+  nvme_cmd.mptr = allocation.mdata_addr;
+  nvme_cmd.dptr.prp.prp1 = allocation.data_addr;
+  nvme_cmd.cdw[0] = lba;              // cdw10 Starting lba bits 31:00
+  nvme_cmd.cdw[2] = transfer_length;  // cdw12 nlb bits 15:00
+
+  return StatusCode::kSuccess;
```
basimsahaf, line 86:
```suggestion
  return status_code;
```

lib/translator/read.cc
```diff
@@ -0,0 +1,187 @@
...
+  status = BuildPrinfo(rdprotect, prinfo);
+
+  if (status != StatusCode::kSuccess) {
+    return status;
+  }
+
+  nvme_cmd.cdw[2] |= (prinfo << 26);  // cdw12 prinfo bits 29:26
+  nvme_cmd.cdw[2] |= (fua << 30);     // cdw12 fua bit 30;
+
+  return StatusCode::kSuccess;
```
basimsahaf, line 109:
```suggestion
  return status_code;
```

lib/translator/read.h
```diff
@@ -0,0 +1,62 @@
...
+// As such, it will call the LegacyRead() translation function
+
+// Read(10), Read(12), and Read(16) have essentially the same fields but with
+// different memory layouts. They all call the Read() translation function,
+// which calls LegacyRead() and handles some additional fields
+
+// Read(32) calls Read() and also translates additional logical block
+// application tags
+
+StatusCode Read6ToNvme(absl::Span<const uint8_t> scsi_cmd,
```
harloff-g, line 40: How do we handle ReadNToScsi?

eyhhuang: not sure what you mean by ReadNToScsi? Each of the Read(6), Read(10), etc. are separate scsi commands with different opcodes

harloff-g: I just mean N as a placeholder. For example, what about Read10ToScsi? How does read data get back to the SCSI data-in buffer?

eyhhuang: Ah, that is to be handled in a separate PR to reduce PR size https://github.com/googleinterns/scsi2nvme/issues/93

basimsahaf: I think once we have a clear direction about the nvme communication, we can go about handling that case

lib/translator/read.cc
basimsahaf: @eyhhuang How are you handling the case where th transfer length is greater than 2^16 multiple? According to the translation spec, in case the transfer length is longer, we need to issue multiple nvm express commands. Refer to section 3.14

basimsahaf: This will be the case with commands Read12/16/32

eyhhuang: Yes I mentioned this in one of the syncs a while ago but Jonathan and I decided that since NVMe has a transfer limit of 16 bits and SCSI sometimes has a limit of 32 bits, it makes sense for us to enforce that the SCSI read commands do not exceed the NVMe device's limits. 
Since we are the SCSI device, we can set our max transfer limits in Block Limits VPD to the NVMe device's max transfer size  (page size * 2^(MDTS)) which will fit within the NVMe protocol limit of 16 bits.

basimsahaf: Okay that makes sense. I'll document this in one of the docs as I have a similar case in `Write` Command

lib/translator/read.cc
eyhhuang: update: Read and Write have different logic when building prinfo, so it is not desirable to abstract this out for both read and write

lib/scsi.h
```diff
@@ -389,11 +389,11 @@ static_assert(sizeof(Read6Command) == 5);
 // SCSI Reference Manual Table 97
 // https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
 struct Read10Command {
-  uint8_t rd_protect : 3;  // read protect bit
-  bool dpo : 1;            // disable page output bit
-  bool fua : 1;            // Forced Unit access bit
-  bool rarc : 1;           // Rebuild Assist Recovery bit
-  bool fua_nv : 1;         // Forced Access Unit Non-Volatile
+  uint8_t rdprotect : 3;  // read protect bit
+  bool dpo : 1;           // disable page output bit
```
toadchild, line 393: Per my comment in the team chat, I'm concerned that these fields are out of order.

lib/translator/read.cc
```diff
@@ -0,0 +1,196 @@
...
+  nvme_cmd = nvme::GenericQueueEntryCmd{
+      .opc = static_cast<uint8_t>(nvme::NvmOpcode::kRead),
+      .psdt = 0,  // PRPs are used for data transfer
+      .nsid = nsid};
+
+  nvme_cmd.mptr = allocation.mdata_addr;
+  nvme_cmd.dptr.prp.prp1 = allocation.data_addr;
+
+  // cdw10 Starting lba bits 31:00
+  nvme_cmd.cdw[0] = static_cast<uint32_t>(lba);
```
toadchild, line 88: Read6 also has a shorter (20 bit) lba field.

toadchild: Withdrawn, I got scsi and nvme structs backwards here.

lib/translator/read.cc
```diff
@@ -0,0 +1,196 @@
...
+  prinfo = prchk | (pract << 3);
+
+  return StatusCode::kSuccess;
+}
+
+// Builds NVMe cdw 12 for Read10, Read12, Read16 translations
+uint32_t BuildCdw12(uint32_t transfer_length, uint8_t prinfo, bool fua) {
+  // cdw12 nlb bits 15:00 (zero based field), printfo bits 29:26, fua bit 30
+  // Since NVMe has a protocol limit of 16 bits on transfer size,
+  // we can enforce a transfer limit of 16 bits on our SCSI device
```
toadchild, line 62: We should still have some sort of error checking to ensure that our length expectation is met.

lib/scsi.h
eyhhuang: Thanks for the catch! This is addressed in https://github.com/googleinterns/scsi2nvme/pull/141

test/test.out
```diff
@@ -0,0 +1,58 @@
+PASSED: //test:tests (see /usr/local/google/home/eyhhuang/.cache/bazel/_bazel_eyhhuang/0bf0c1097a38e6ad1fca7c8fe1b81615/execroot/__main__/bazel-out/k8-fastbuild/testlogs/test/tests/test.log)
```
Ralith, line 1: Is this file included intentionally?

eyhhuang: oh oops, i'll remove that!

lib/translator/read.cc
```diff
@@ -0,0 +1,213 @@
...
+    return StatusCode::kInvalidInput;
+  }
+  // Transfer Length set to 0 specifies 256 logical blocks to be read
+  // Section 3.15 Seagate SCSI specs
+  uint16_t updated_transfer_length =
+      read_cmd.transfer_length == 0 ? 256 : read_cmd.transfer_length;
+
+  // Transform logical_block_address to network endian before casting to 64 bits
+  StatusCode status = LegacyRead(
+      htonl(read_cmd.logical_block_address), nvme_cmd, allocation, nsid,
```
toadchild, line 152: I think that read6 lba now requires some bit math to reassemble.

lib/translator/read.cc
```diff
@@ -109,26 +103,24 @@ StatusCode Read(uint8_t rd_protect, bool fua, uint64_t lba,
...
 
   uint8_t prinfo;  // Protection Information field 4 bits;
   status = BuildPrinfo(rd_protect, prinfo);
-
   if (status != StatusCode::kSuccess) {
     return status;
   }
 
-  nvme_cmd.cdw[2] = BuildCdw12(transfer_length, prinfo, fua);
+  nvme_cmd.cdw[2] = __bswap_32(htonl(BuildCdw12(transfer_length, prinfo, fua)));
```
toadchild, line 121: If you're going to do a double byteswap to guarantee that the variable is little endian, please define a function to do so.

lib/translator/read.cc
```diff
@@ -147,19 +139,22 @@ StatusCode Read6ToNvme(Span<const uint8_t> scsi_cmd,
...
+  // cdw10 Starting lba bits 31:00
+  uint32_t host_endian_lba = (read_cmd.logical_block_address_1 << 16) +
+                             ntohs(read_cmd.logical_block_address_2);
+  nvme_cmd.cdw[0] = __bswap_32(htonl(host_endian_lba));
 
-  // cdw12 nlb bits 15:00 (zero based field)
-  nvme_cmd.cdw[2] = updated_transfer_length - 1;
+  // nlb is a zero-based field
+  // cdw12 nlb bits 15:00
+  nvme_cmd.cdw[2] = __bswap_16(htons(updated_transfer_length - 1));
```
toadchild, line 155: Is cdw[] defined as 32 or 16 bit?  Assigning a 16 bit value to a 32 bit variable will have the compiler zero-pad the value based on host byte order.

lib/translator/read.cc
eyhhuang: good idea, defined helper functions here https://github.com/googleinterns/scsi2nvme/pull/157

lib/translator/read.cc
eyhhuang: good catch, it is 32 bit

lib/translator/read.cc
toadchild: Looks good!

lib/translator/read.cc
```diff
@@ -0,0 +1,232 @@
...
+                       uint32_t page_size, uint32_t lba_size) {
+  scsi::Read6Command read_cmd;
+  if (!ReadValue(scsi_cmd, read_cmd)) {
+    DebugLog("Malformed Read6 command");
+    return StatusCode::kInvalidInput;
+  }
+  // Transfer Length set to 0 specifies 256 logical blocks to be read
+  // Section 3.15 Seagate SCSI specs
+  uint16_t updated_transfer_length =
+      read_cmd.transfer_length == 0 ? 256 : read_cmd.transfer_length;
```
toadchild, line 140: Convert to host byte order here

lib/translator/read.cc
```diff
@@ -0,0 +1,232 @@
...
+
+  StatusCode status = LegacyRead(
+      nvme_cmd, allocation, nsid,
+      GetTransferLengthPages(updated_transfer_length, page_size, lba_size));
+  if (status != StatusCode::kSuccess) {
+    return status;
+  }
+  // cdw10 Starting lba bits 31:00
+  uint32_t host_endian_lba = (read_cmd.logical_block_address_1 << 16) +
```
toadchild, line 149: You're byte swapping lba_2 but not lba_1

toadchild: Never mind - it's because lba_1 is only one byte.  I keep forgetting read6 has a completely ridiculous 24 bit lba field.

lib/translator/read.cc
eyhhuang: read_cmd.transfer_length is actually a single byte. I'll add a static cast so it's clearer

PR #100, opened 2020-07-06
Translate Log Sense - Supported Pages List

JustinAWei:
NOTE: This command is not prioritized and will probably not be worked on any time soon. 

The Log Sense data returns extra information about the device such as Temperature. This PR was not deleted in case Log Sense is picked up again in the future.

### DIFF
--- a/lib/scsi_defs.h
+++ b/lib/scsi_defs.h
@@ -898,6 +898,15 @@ struct LogSenseCommand {
 } ABSL_ATTRIBUTE_PACKED;
 static_assert(sizeof(LogSenseCommand) == 10);
 
+struct SupoprtedLogPages {
+  bool ds :1;
+  bool spf_1b :1;
+  PageCode page_code : 6;
+  uint8_t _subpage_code : 8;
+  uint16_t page_len : 16;
+} ABSL_ATTRIBUTE_PACKED;
+static_assert(sizeof(SupoprtedLogPages) == 4);
+
 }  // namespace scsi_defs
 
 #endif
--- a/lib/scsi_defs.h
+++ b/lib/scsi_defs.h
@@ -27,9 +27,9 @@ using LunAddress = uint64_t;
 
 enum class PageCode : uint8_t {
   kSupportedLogPages = 0x00,
-  Temperature = 0x0d,
-  SolidStateMedia = 0x11,
-  InformationalExceptions = 0x2f
+  kTemperature = 0x0d,
+  kSolidStateMedia = 0x11,
+  kInformationalExceptions = 0x2f
 };
 
 // SAM-4 Table 33
--- a/lib/scsi_defs.h
+++ b/lib/scsi_defs.h
@@ -898,14 +898,14 @@ struct LogSenseCommand {
 } ABSL_ATTRIBUTE_PACKED;
 static_assert(sizeof(LogSenseCommand) == 10);
 
-struct SupoprtedLogPages {
+struct SupportedLogPages {
   bool ds :1;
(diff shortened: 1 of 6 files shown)


lib/translator/logsense.cc
```diff
@@ -0,0 +1,57 @@
...
+      scsi_defs::PageCode::kInformationalExceptions};
+  memcpy(buffer.data(), &result, sizeof(scsi_defs::SupportedLogPages));
+  memcpy(&buffer[sizeof(scsi_defs::SupportedLogPages)], supported_pages_list,
+         sizeof(supported_pages_list));
+}
+
+// Main logic engine for the Inquiry command
+void translate(const scsi_defs::LogSenseCommand &cmd,
+               absl::Span<uint8_t> buffer) {
+  if (cmd.sp == 1 || cmd.pc == 1 || cmd.control_byte.naca == 1) {
```
toadchild, line 33: Is there a particular action that should be taken if any of these are / aren't set?

lib/scsi_defs.h
```diff
@@ -25,6 +25,13 @@ namespace scsi_defs {
 
 using LunAddress = uint64_t;
 
+enum class PageCode : uint8_t {
```
harloff-g, line 28: There's a number of page code types in SCSI. It would be more organized/readable to call this LogPageCode

lib/translator/BUILD
```diff
@@ -5,3 +5,18 @@ cc_library(
...
+  deps = [
+      "//lib/translator:common",
+      "//third_party/spdk_defs:nvme_defs_lib",
+      "//lib:scsi_defs_lib",
+      "@com_google_absl//absl/base",
+      "@com_google_absl//absl/types:span",
+      "@com_google_absl//absl/strings",
+  ],
+  visibility = ["//visibility:public"],
+)
```
harloff-g, line 22: nit: newline

lib/translator/logsense.cc
```diff
@@ -0,0 +1,57 @@
...
+
+namespace translator {
+
+void TranslateSupportedLogPages(absl::Span<uint8_t> buffer) {
+  scsi_defs::SupportedLogPages result = {.page_len = 0x4};
+  scsi_defs::PageCode supported_pages_list[4] = {
+      scsi_defs::PageCode::kSupportedLogPages,
+      scsi_defs::PageCode::kTemperature, scsi_defs::PageCode::kSolidStateMedia,
+      scsi_defs::PageCode::kInformationalExceptions};
+  memcpy(buffer.data(), &result, sizeof(scsi_defs::SupportedLogPages));
```
harloff-g, line 25: #104

lib/translator/logsense.h
```diff
@@ -0,0 +1,33 @@
...
+#define LIB_TRANSLATOR_LOGSENSE_H
+
+#include "absl/strings/string_view.h"
+#include "absl/types/span.h"
+#include "common.h"
+#include "lib/scsi_defs.h"
+#include "third_party/spdk_defs/nvme_defs.h"
+
+namespace translator {
+void translate(const scsi_defs::LogSenseCommand &cmd, absl::Span<uint8_t>);
```
harloff-g, line 25: This PR doesn't seem to follow the pattern other PRs have been using. That is, the header exposes a "CommandToNvme" and "CommandToScsi" function. All other functions are part of an anonymous namespace

lib/translator/logsense.cc
JustinAWei: we would return one of them scsi codes which is handled in the translator api

PR #105, opened 2020-07-06
Memory Tracking/Cleanup

harloff-g:
This PR defines the interface through which the translation object tracks and deallocates memory.

Recent work on communicating with the NVMe driver performed by Basim and Michal has revealed that the passback-memory the library has to keep track of are simply pointers to large buffers of memory (not PRPs). See [this](https://github.com/linux-nvme/nvme-cli/blob/master/nvme.c#L90) code for relevant examples. 

This means that we can track all memory allocated by the library using a rigidly defined structure which holds two pointers, one for a data buffer and one for a metadata buffer. These two pointers cover any given NVMe command's allocation needs.


lib/translator/common.h
```diff
@@ -44,6 +46,13 @@ void DeallocPages(void* pages_ptr, uint16_t count);
 void SetAllocPageCallbacks(void* (*alloc_callback)(uint16_t),
                            void (*dealloc_callback)(void*, uint16_t));
 
+struct allocation {
```
Ralith, line 49: Type names should be capitalized.

lib/translator/common.h
```diff
@@ -44,6 +46,13 @@ void DeallocPages(void* pages_ptr, uint16_t count);
...
+  void* data_addr;      // start of data buffer
+  uint16_t data_size;   // data page size
+  void* mdata_addr;     // start of metadata buffer
+  uint16_t mdata_size;  // metadata page size
```
Ralith, line 53: I'm a bit confused by the notion of "page size" here. This may be because I'm lacking some context, but if so some more comments might be justified:

Aren't pages more or less fixed-size? Are these the size of one page, a quantity of pages, or the size of a single allocation that may span multiple pages? If the last, why even describe them in terms of pages, since that's just standard memory allocation?

harloff-g: Poor word choice on my part. These are intended to capture the number of pages that were allocated for the NVMe buffer. Pages were chosen over standard memory allocation because NVMe responses are almost always multiples of pages. I'll update with clearer phrasing!

Ralith: Thanks, that's clearer!

lib/translator/translation.cc
```diff
@@ -75,7 +77,16 @@ absl::Span<const nvme_defs::GenericQueueEntryCmd> Translation::GetNvmeCmds() {
 void Translation::AbortPipeline() {
   pipeline_status_ = StatusCode::kUninitialized;
   nvme_cmd_count_ = 0;
-  // TODO free allocated PRPs
+  FlushMemory();
+}
+
+void Translation::FlushMemory() {
+  for (int i = 0; i < kMaxCommandRatio; ++i) {
```
Ralith, line 84: Should this only iterate to `nvme_cmd_count_`?

lib/translator/translation.cc
```diff
@@ -75,7 +77,16 @@ absl::Span<const nvme_defs::GenericQueueEntryCmd> Translation::GetNvmeCmds() {
...
+    if (allocations_[i].data_addr != nullptr)
+      DeallocPages(allocations_[i].data_addr, allocations_[i].data_page_count);
+    if (allocations_[i].mdata_addr != nullptr)
+      DeallocPages(allocations_[i].mdata_addr, allocations_[i].mdata_page_count);
```
Ralith: We should set the addrs to `nullptr` afterwards so a reuse of the object doesn't cause double-frees. Similarly, `Translation` will now need an explicit constructor that sets the addrs to `nullptr` when the object is first created.

lib/translator/translation.cc
harloff-g: I had it at kMaxCommandRatio just as a safety precaution in case a translation function skipped an index but as long as it's expected no indexes are skipped then using nvme_cmd_count_ is appropriate

lib/translator/common.h
```diff
@@ -44,13 +46,28 @@ void DeallocPages(void* pages_ptr, uint16_t count);
...
+};
+
 struct BeginResponse {
   ApiStatus status;
   uint32_t alloc_len;
 };
 
 class Translation {
  public:
+  Translation() {
```
JustinAWei, line 63: perhaps initializer list would be cleaner here
`Translation() : allocations_({nullptr})`

Ralith: The `{nullptr}` here is unnecessary; just `allocations_()` is sufficient to force zero-initialization. See https://en.cppreference.com/w/cpp/language/zero_initialization.

lib/translator/common.h
```diff
@@ -20,6 +20,8 @@
 
 namespace translator {
 
+constexpr int kMaxCommandRatio = 3;
```
Ralith, line 23: Could use a few words of documentation.

lib/translator/translation.h
```diff
@@ -29,6 +29,10 @@ struct BeginResponse {
 
 class Translation {
  public:
+  Translation()
+      : pipeline_status_(StatusCode::kUninitialized),
+        nvme_cmd_count_(),
```
Ralith, line 34: While not strictly necessary, it's usually clearer to use an explicit 0 when initializing primitive values. We do differently for arrays just because they're much more awkward to initialize any other way.

Issue #104, opened 2020-07-06
Create templated function to perform checked memcpy


eyhhuang:
Check the size of data buffer before copying it to the desired target.

PR #106, opened 2020-07-06
Templated memcpy

eyhhuang:
Closes https://github.com/googleinterns/scsi2nvme/issues/104


lib/translator/common.h
```diff
@@ -23,6 +27,14 @@ void DebugLog(const char* format, ...);
 
 void SetDebugCallback(void (*callback)(const char*));
 
+template <typename T>
+bool ReadValue(absl::Span<const uint8_t> data, T& out) {
+  static_assert(std::is_pod_v<T>, "only supports POD types");
```
Ralith, line 32: The `type_traits` header should be included explicitly for `is_pod_v`, even though it happens to work anyway by coincidence for now.

test/translator/common_test.cc
```diff
@@ -34,4 +36,30 @@ TEST(Common, ShouldCorrectlyCallback) {
   translator::DebugLog(buf, 123);
 }
 
+TEST(Common, ShouldNotMemcpy) {
+  scsi_defs::Read6Command cmd;
+  uint8_t buffer[1];
+
+  auto span = absl::MakeSpan(buffer, sizeof(scsi_defs::Read6Command) + 1);
```
Ralith, line 43: I'm uncomfortable constructing spans like this that address invalid memory. Just make `buffer` big enough, maybe?

lib/translator/common.h
```diff
@@ -23,6 +27,14 @@ void DebugLog(const char* format, ...);
...
+  if (sizeof(T) != data.size()) return false;
```
harloff-g, line 33: I think we might find it useful in more scenarios to just copy as much data as possible into the structure "out" rather than enforcing the span be exactly of the same size. For example, imagine we have a large span but we know that the next 10 bytes is a particular structure that we want to cast.
So, if (sizeof(T) > data.size()) copySize = data.size();

Ralith: It's really easy to create a span for a prefix of another span (`foo.subspan(0, n)`), but I'm not sure requiring that boilerplate from callers adds any value, so I agree that being permissive here is sensible.

eyhhuang: That makes sense to me!

eyhhuang:
> Do we need the complimentary operation for writing values out to a mutable span, as well?

Initially I didn't think that we needed a WriteValue, but now I realize we probably need it for writing the SCSI response to the buffer

lib/translator/common.h
```diff
@@ -23,6 +28,22 @@ void DebugLog(const char* format, ...);
...
+  static_assert(std::is_pod_v<T>, "Only supports POD types");
+  if (sizeof(T) > data.size()) return false;
+  memcpy(&out, data.data(), sizeof(T));
+  return true;
+}
+
+template <typename T>
+bool WriteValue(const T& data, absl::Span<uint8_t> out) {
+  static_assert(std::is_pod_v<T>, "Only supports POD types");
+  if (out.size() > sizeof(T)) return false;
```
harloff-g, line 42: Isn't the goal of this function simply to write the type into the span? If so, we would have: 
memcpy(out.data(), &data, sizeof(T)); 
and 
if (sizeof(T) > data.size()) return false;

eyhhuang: oops, you're right. in this case we would want to ensure we write the entire T obj.

PR #108, opened 2020-07-08
ScsiOpcodeToString helper

JustinAWei:


lib/translator/common.h
```diff
@@ -28,6 +30,8 @@ void DebugLog(const char* format, ...);
 
 void SetDebugCallback(void (*callback)(const char*));
 
+absl::string_view ScsiOpcodeToString(scsi_defs::OpCode opcode);
```
Ralith, line 33: It might be difficult to use this with `DebugLog` since afaik there's no way to use a `string_view` with the printf family of functions, and we don't want to heap-alloc a `std::string`. If that was the intention, it might be best to return a `const char*` here contrary to usual good style, just so we can keep `DebugLog` simple.

PR #115, opened 2020-07-09
rm _defs keyword

JustinAWei:


lib/translator/inquiry.cc
```diff
@@ -0,0 +1,252 @@
+// Copyright 2020 Google LLC
```
harloff-g, line 1: Is this file mistakenly included in this PR?

JustinAWei: ah yep, leftover from editor

PR #116, opened 2020-07-10
Add Allocation::SetPages() helper fn

eyhhuang:
`Allocation::SetPages(uint16_t data_page_count, uint16_t mdata_page_count)` sets the page counts of the allocation object, calls `AllocPages` and returns a status code depending on whether `AllocPages` was successful


test/translator/common_test.cc
```diff
@@ -78,4 +78,40 @@ TEST(Common, ShouldCorrectlyWriteValueToSpan) {
...
+TEST(Common, ShouldBuildAllocationWithSuccessStatus) {
+  translator::Allocation allocation;
+  auto alloc_callback = [](uint16_t count) -> uint64_t {
+    EXPECT_EQ(1, count);
+    return 1337;
+  };
+  void (*dealloc_callback)(uint64_t, uint16_t) = nullptr;
+  translator::SetAllocPageCallbacks(alloc_callback, dealloc_callback);
+
+  translator::StatusCode status_code = allocation.SetPages(1, 1);
```
toadchild, line 90: I would recommend passing and returning different values for data and mdata so that you can verify that the right values made it into each field.

eyhhuang: That's a good idea, will do!

lib/translator/common.cc
```diff
@@ -54,6 +54,21 @@ void SetAllocPageCallbacks(uint64_t (*alloc_callback)(uint16_t),
...
 }
 
+StatusCode Allocation::SetPages(uint16_t data_page_count,
+                                uint16_t mdata_page_count) {
+  this->data_page_count = data_page_count;
+  this->data_addr = AllocPages(data_page_count);
+  this->mdata_page_count = mdata_page_count;
+  this->mdata_addr = AllocPages(mdata_page_count);
+
+  if (this->data_addr == 0 || this->mdata_addr == 0) {
```
harloff-g, line 64: What if mdata_page_count is set to 0 and the desired behavior is for the mdata_addr to be 0? We probably need a more "conditional" approach to this

eyhhuang: that's a good point. Should I also account for the case where data_page_count is set to 0? Or is that not a meaningful situation?

harloff-g: I think it could potentially be a meaningful situation; however, I can't think of any cases off the top of my head that would require that functionality. Nonetheless, it's probably good practice to handle both cases synonymously

lib/translator/common.cc
```diff
@@ -54,6 +54,22 @@ void SetAllocPageCallbacks(uint64_t (*alloc_callback)(uint16_t),
...
+                                uint16_t mdata_page_count) {
+  this->data_page_count = data_page_count;
+  this->data_addr = AllocPages(data_page_count);
+  this->mdata_page_count = mdata_page_count;
+  this->mdata_addr = AllocPages(mdata_page_count);
+
+  if (this->data_addr == 0 ||
+      (this->mdata_addr == 0 && mdata_page_count != 0)) {
+    DebugLog("Error when requesting memory");
+    return StatusCode::kFailure;
```
Ralith, line 67: Would it make sense to free the successful allocation, if only one failed?

test/translator/common_test.cc
```diff
@@ -78,4 +78,65 @@ TEST(Common, ShouldCorrectlyWriteValueToSpan) {
...
+    if (count == 1) {
+      return 1337;
+    }
+    if (count == 3) {
+      return 7331;
+    }
+    return 0;
+  };
+  void (*dealloc_callback)(uint64_t, uint16_t) = nullptr;
+  translator::SetAllocPageCallbacks(alloc_callback, dealloc_callback);
```
Ralith, line 93: Is there any mechanism preventing race conditions when setting this in multiple tests?

lib/translator/common.cc
eyhhuang: I believe there is a PR to call `FlushMemory` for anything that has been allocated even on failure, so I don't think we need to do any extra memory handling

test/translator/common_test.cc
eyhhuang: There currently isn't a mechanism for this, but I could definitely see value in that to prevent flakiness. Do you have any suggestions?

Ralith: On review it looks like gtest doesn't have any built-in parallelism, and environments that apply parallelism to it do so via multiple processes. so this probably isn't an issue after all.

lib/translator/common.h
```diff
@@ -43,6 +43,10 @@ struct Allocation {
   uint16_t data_page_count;   // Number of data pages
   uint64_t mdata_addr;        // Start of metadata buffer
   uint16_t mdata_page_count;  // Number of metadata pages
+
+  // Sets [m]data_page_count, calls AllocPages([m]data_page_count),
+  // and returns StatusCode based on whether AllocPages was successful
+  StatusCode SetPages(uint16_t data_page_count, uint16_t mdata_page_count);
```
harloff-g, line 49: We could implement a safety check to prevent a call to SetPages from overriding page data that has not yet been flushed

Issue #117, opened 2020-07-10
Translate TestUnitReady


JustinAWei:
TEST UNIT READY support requires a status to be returned to application client indicating the status of
the logical unit. If the NVM Express device is ready to accept commands a status of GOOD shall be
returned, otherwise a status of CHECK CONDITION and a sense key of NOT READY shall be returned.

Controller registers are located in the MLBAR/MUBAR registers (PCI BAR0 and BAR1) that shall be
mapped to a memory space that supports in-order access and variable access widths.

From offset 14h in the Controller registers, the Enable bit should indicate whether the NVMe device is ready or not.
"When set to ‘1’, then the controller shall process commands based on Submission Queue Tail doorbell writes"

PR #120, opened 2020-07-10
Translate TestUnitReady

JustinAWei:
This relies on the NVMe side providing an abstraction to check NVMe device ready. See #117

### DIFF
--- a/lib/translator/test_unit_ready.cc
+++ b/lib/translator/test_unit_ready.cc
@@ -13,11 +13,26 @@
 // limitations under the License.
 
 #include "test_unit_ready.h"
-
-#include "common.h"
-
 namespace translator {
+namespace {
+// TODO: mocked. this fn should be on the NVMe side.
+StatusCode NvmeReady(bool& result) {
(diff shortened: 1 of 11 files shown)


lib/translator/test_unit_ready.cc
```diff
@@ -0,0 +1,41 @@
...
+namespace {
+// TODO: mocked. this fn should be on the NVMe side.
+StatusCode NvmeReady(bool& result) {
+  result = true;
+  return StatusCode::kSuccess;
+}
+
+}  // namespace
+
+StatusCode TestUnitReadyToNvme(absl::Span<const uint8_t> scsi_cmd) {
```
eyhhuang, line 26: I think this should be renamed something like ValidateTestUnitReady rather than ..ToNvme because this doesn't actually do any translation. Or maybe we can remove this function altogether

toadchild: Please add a comment explaining what's going on in this function, since it's very different than most of the other translations.

PR #124, opened 2020-07-13
Translate Read Capacity 10

JustinAWei:
#122


lib/translator/BUILD
```diff
@@ -1,3 +1,17 @@
+cc_library(
+  name = "read_capacity_10_lib",
```
harloff-g, line 2: Better to have  "//lib/translator:common" as a dependency than to include common.h and .cc as part of the headers and source

lib/translator/read_capacity_10.cc
```diff
@@ -0,0 +1,70 @@
...
+                                nvme::GenericQueueEntryCmd& identify_ns,
+                                scsi::LunAddress lun) {
+  scsi::ReadCapacity10Command cmd = {};
+  if (!ReadValue(raw_scsi, cmd)) {
+    DebugLog("Malformed ReadCapacity10 Command");
+    return StatusCode::kInvalidInput;
+  };
+
+  if (cmd.control_byte.naca == 1) {
+    DebugLog("Malformed ReadCapacity10 Command");
```
harloff-g, line 31: Might be useful to print a slightly more detailed/different error message here so it's easier to tell where things are going wrong when debugging

lib/translator/read_capacity_10.cc
```diff
@@ -0,0 +1,70 @@
...
+    return StatusCode::kInvalidInput;
+  }
+
+  // Allocate prp & assign to command
+  uint64_t ns_prp = AllocPages(1);
```
eyhhuang, line 36: Should be using the Allocation object

lib/translator/read_capacity_10.h
```diff
@@ -0,0 +1,37 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIB_TRANSLATOR_read_capacity_10_H
+#define LIB_TRANSLATOR_read_capacity_10_H
+
+#include <cstdio>
```
eyhhuang, line 18: is this needed?

lib/translator/read_capacity_10.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include "common.h"
+#include "lib/scsi.h"
```
eyhhuang, line 21: our own project's .h files should be under third party ones

lib/translator/translation.cc
```diff
@@ -69,10 +74,15 @@ ApiStatus Translation::Complete(
   // Switch cases should not return
   ApiStatus ret;
   scsi::OpCode opc = static_cast<scsi::OpCode>(scsi_cmd_[0]);
+
   switch (opc) {
     case scsi::OpCode::kInquiry:
       ret = ApiStatus::kSuccess;
       break;
+    case scsi::OpCode::kReadCapacity10:
+      pipeline_status_ = ReadCapacity10ToScsi(buffer, nvme_cmds_);
```
eyhhuang, line 83: nit: i think you could just pass in nvme_cmds_[0]

test/translator/read_capacity_10_test.cc
```diff
@@ -0,0 +1,166 @@
...
+            translator::ReadCapacity10ToScsi(buffer_, nvme_cmds_));
+  scsi::ReadCapacity10Data result = {};
+  EXPECT_TRUE(translator::ReadValue(buffer_, result));
+  EXPECT_EQ(result.returned_logical_block_address,
+            identify_ns_.nsze > 0xffffffff ? 0xffffffff : identify_ns_.nsze);
+  EXPECT_EQ(result.block_length,
+            identify_ns_.lbaf[identify_ns_.flbas.format].lbads);
+}
+
+}  // namespace
```
eyhhuang, line 166: nit: new line at end of file

test/translator/read_capacity_10_test.cc
```diff
@@ -0,0 +1,170 @@
...
+  nvme::GenericQueueEntryCmd identify_cmds_[1];
+  absl::Span<nvme::GenericQueueEntryCmd> nvme_cmds_;
+  absl::Span<const uint8_t> scsi_cmd_;
+  nvme::IdentifyNamespace identify_ns_;
+
+  uint8_t buffer_[200];
+
+  // Per-test-suite set-up.
+  // Called before the first test in this test suite.
+  static void SetUpTestSuite() {
```
Ralith, line 32: Nice, I didn't know this existed!

test/translator/read_capacity_10_test.cc
```diff
@@ -0,0 +1,170 @@
...
+      // TODO: fix count == 0
+      // EXPECT_EQ(1, count);
+      return 1337;
+    };
+    void (*dealloc_callback)(uint64_t, uint16_t) = nullptr;
+    translator::SetAllocPageCallbacks(alloc_callback, dealloc_callback);
+  }
+
+  // You can define per-test set-up logic as usual.
+  virtual void SetUp() {
```
Ralith, line 43:
```suggestion
  void SetUp() override {
```
Methods should only be marked `virtual` if some other class is later going to override them. Conversely, if you're intending to override a method, you should mark it `override`; this both clarifies your intent to a reader, and allows the compiler to warn you if for whatever reason you aren't actually overriding anything.

test/translator/read_capacity_10_test.cc
```diff
@@ -0,0 +1,170 @@
...
+}
+
+TEST_F(ReadCapacity10Test, Success) {
+  identify_ns_.nsze = 0;
+  identify_ns_.flbas.format = 0;
+  identify_ns_.lbaf[identify_ns_.flbas.format].lbads = 0;
+  EXPECT_EQ(translator::StatusCode::kSuccess,
+            translator::ReadCapacity10ToScsi(buffer_, nvme_cmds_[0]));
+  scsi::ReadCapacity10Data result = {};
+  EXPECT_TRUE(translator::ReadValue(buffer_, result));
```
Ralith, line 94: Use `ASSERT` rather than `EXPECT` when it doesn't make sense to continue the test after a failure, as is the case here and in other uses of `ReadValue` below.

lib/translator/BUILD
```diff
@@ -1,3 +1,16 @@
...
```
toadchild, line 2: Are we planning on supporting ReadCapacity16?  Any reason to not make a single `read_capacity_lib` and have both 10 and 16 in the same file? (This is the pattern that Read and Write appear to be following)

lib/translator/read_capacity_10.cc
```diff
@@ -0,0 +1,69 @@
...
+  // Allocate prp & assign to command
+  auto read_status = allocation.SetPages(1, 0);
+  if (read_status != StatusCode::kSuccess) {
+    return read_status;
+  }
+
+  identify_ns = nvme::GenericQueueEntryCmd{
+      .opc = static_cast<uint8_t>(nvme::AdminOpcode::kIdentify), .nsid = lun};
+  identify_ns.dptr.prp.prp1 = allocation.data_addr;
+  identify_ns.cdw[0] = 0x0;  // cns val
```
oicirtap, line 42: What does cns stand for?

lib/translator/read_capacity_10.cc
```diff
@@ -0,0 +1,69 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "read_capacity_10.h"
+
+namespace translator {
+
+StatusCode ReadCapacity10ToNvme(absl::Span<const uint8_t> raw_scsi,
+                                nvme::GenericQueueEntryCmd& identify_ns,
+                                scsi::LunAddress lun, Allocation allocation) {
```
oicirtap, line 21: Do we have documentation on how `Allocation` is used, and how it related to `identify_ns`? It's unclear to me what we use the allocation object for.

lib/translator/read_capacity_10.cc
```diff
@@ -0,0 +1,69 @@
...
+  identify_ns = nvme::GenericQueueEntryCmd{
+      .opc = static_cast<uint8_t>(nvme::AdminOpcode::kIdentify), .nsid = lun};
+  identify_ns.dptr.prp.prp1 = allocation.data_addr;
+  identify_ns.cdw[0] = 0x0;  // cns val
+
+  return StatusCode::kSuccess;
+}
+
+StatusCode ReadCapacity10ToScsi(absl::Span<uint8_t> buffer,
+                                nvme::GenericQueueEntryCmd gen_identify_ns) {
```
oicirtap, line 48: Why do we need an entire GenericQueueEntryCmd as input? Is this a convention we are using for all `ToScsi` functions? Would it be cleaner to pass as an input the value of `gen_identify_ns.dptr.prp.prp1` instead?

lib/translator/read_capacity_10.cc
```diff
@@ -0,0 +1,69 @@
...
+  uint8_t* ns_dptr = reinterpret_cast<uint8_t*>(gen_identify_ns.dptr.prp.prp1);
+  auto ns_span = absl::MakeSpan(ns_dptr, sizeof(nvme::IdentifyNamespace));
+
+  nvme::IdentifyNamespace identify_ns = {};
+  if (!ReadValue(ns_span, identify_ns)) {
+    DebugLog("Malformed IdentifyNamespace Data");
+    return StatusCode::kInvalidInput;
+  };
+
+  uint8_t lba_format_index = identify_ns.flbas.format;
```
oicirtap, line 58: It's probably fine to inline this and remove the `lba_format_index `. To me personally the extra variable doesn't improve readability, but I'll leave it up to you.

lib/translator/read_capacity_10.cc
eyhhuang: for quick context, `Allocation` is a way to store and manage all the memory that is allocated so we can properly deallocate it at the end each translation. See this [PR](https://github.com/googleinterns/scsi2nvme/pull/105/files#diff-e8eb0eb4e8ba5e2a5b4c9411b3673f7dR41). I'm working to add more documentation in our `SCSI2NVMe Translation Spec`

lib/translator/BUILD
JustinAWei: ReadCapacity10 was specifically specified for minimum core functionality by @w-michal. When we add ReadCapacity16, it does make sense to add it to this file.

lib/translator/read_capacity_10.cc
JustinAWei: this is currently a convention for our generic translator. we can create a separate issue to change this standard if needed

test/translator/read_capacity_10_test.cc
```diff
@@ -0,0 +1,170 @@
...
+    auto alloc_callback = [](uint16_t count) -> uint64_t {
+      // TODO: fix count == 0
+      // EXPECT_EQ(1, count);
+      return 1337;
+    };
+    void (*dealloc_callback)(uint64_t, uint16_t) = nullptr;
+    translator::SetAllocPageCallbacks(alloc_callback, dealloc_callback);
+  }
+
+  // You can define per-test set-up logic as usual.
```
Ralith, line 42: Is this commit a stray copy-paste?

JustinAWei: yep but it is also correctly descriptive

PR #125, opened 2020-07-13
Translate Mode Sense

### LINKED ISSUE #73: Translate Mode Sense Command
Devise a plan for translating SCSI Mode Sense (6 and 10) to NVMe equivalent. Implement solution in SCSI2NVMe translation library.

harloff-g:
Closes #73 
Code to translate mode sense 6 and mode sense 10


lib/translator/translation.cc
```diff
@@ -36,18 +37,26 @@ BeginResponse Translation::Begin(absl::Span<const uint8_t> scsi_cmd,
   pipeline_status_ = StatusCode::kSuccess;
   scsi_cmd_ = scsi_cmd;
 
+  uint32_t nsid = static_cast<uint32_t>(lun) + 1;
```
eyhhuang, line 40: why is this + 1

lib/translator/mode_sense.cc
```diff
@@ -0,0 +1,390 @@
...
+  bool dbd;
+  bool llbaa;
+};
+
+// Generates an nvme identify command for getting block descriptor data
+StatusCode GenerateBlockDescriptorIdentifyCmd(
+    nvme::GenericQueueEntryCmd& nvme_cmd, Allocation& allocation,
+    uint32_t nsid) {
+  nvme_cmd = {.opc = static_cast<uint8_t>(nvme::AdminOpcode::kIdentify),
+              .nsid = nsid};
```
eyhhuang, line 77: should the nsid be byteswapped for endianness

test/translator/mode_sense_test.cc
```diff
@@ -0,0 +1,268 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "lib/translator/mode_sense.h"
+
+#include "gtest/gtest.h"
+
+#include <iostream>
```
eyhhuang, line 19: this can be removed

lib/translator/mode_sense.cc
```diff
@@ -0,0 +1,390 @@
...
+  uint64_t prp_ptr = AllocPages(1);
+  if (prp_ptr == 0) return StatusCode::kFailure;
+  allocation.data_addr = prp_ptr;
+  allocation.data_page_count = 1;
+  nvme_cmd.dptr.prp.prp1 = prp_ptr;
```
eyhhuang, line 82: https://github.com/googleinterns/scsi2nvme/pull/116 has been merged so the SetPages can be used

lib/translator/mode_sense.cc
```diff
@@ -0,0 +1,390 @@
...
+  nvme::IdentifyNamespace idns;
+  uint8_t* prp_ptr = reinterpret_cast<uint8_t*>(identify.dptr.prp.prp1);
+  ReadValue(absl::MakeSpan(prp_ptr, sizeof(nvme::IdentifyNamespace)), idns);
+  if (llbaa) {
+    write_len = sizeof(scsi::LongLbaBlockDescriptor);
+    scsi::LongLbaBlockDescriptor lbd = {};
+    lbd.number_of_blocks = idns.ncap;
+    // Figure 246
+    // https://nvmexpress.org/wp-content/uploads/NVM-Express-1_4-2019.06.10-Ratified.pdf
+    lbd.logical_block_length = 2 ^ (idns.lbaf[idns.flbas.format].lbads);
```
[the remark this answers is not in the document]
harloff-g: "This field indicates the LBA data size supported. The value is reported in terms of a power of two (2^n)."

lib/translator/translation.cc
harloff-g: Lun => nsid mapping has an offset of one because NSIDs start at 1 and LUNs are required to start at 0. I put this here so that going forward individual functions don't have to worry about LUN => nsid transformation and can just rely on this provided value.

lib/translator/mode_sense.cc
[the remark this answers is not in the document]
harloff-g: Good catch, updating to use bit shift

lib/translator/mode_sense.cc
harloff-g: I think we will be able to ignore the endianness of the LUN as it is stored in a memory field and not buffer https://elixir.bootlin.com/linux/v4.2/source/include/scsi/scsi_device.h#L102

lib/translator/BUILD
```diff
@@ -65,3 +66,13 @@ cc_library(
   ],
   visibility = ["//visibility:public"],
 )
+
+cc_library(
+  name = "mode_sense_lib",
+  hdrs = ["mode_sense.h"],
+  srcs = ["mode_sense.cc"],
+  deps = [
+    "//lib/translator:common"
```
JustinAWei, line 75: :

lib/translator/mode_sense.cc
```diff
@@ -0,0 +1,389 @@
...
+      break;
+    case scsi::PageControl::kDefault:
+      tmp_get_features.sel = nvme::FeatureSelect::kDefault;
+      break;
+    default:
+      DebugLog("Unsupported page control recieved");
+      return StatusCode::kFailure;
+  }
+  tmp_get_features.fid = nvme::FeatureType::kVolatileWriteCache;
+  memcpy(&nvme_cmd, &tmp_get_features, sizeof(nvme::GenericQueueEntryCmd));
```
JustinAWei, line 105: readval

lib/translator/translation.cc
```diff
@@ -77,13 +91,21 @@ ApiStatus Translation::Complete(
...
 
   // Switch cases should not return
-  ApiStatus ret;
+  ApiStatus ret = ApiStatus::kSuccess;
   absl::Span<const uint8_t> scsi_cmd_no_op = scsi_cmd_.subspan(1);
   scsi::OpCode opc = static_cast<scsi::OpCode>(scsi_cmd_[0]);
   switch (opc) {
     case scsi::OpCode::kInquiry:
       pipeline_status_ = InquiryToScsi(scsi_cmd_no_op, buffer, GetNvmeCmds());
-      ret = ApiStatus::kSuccess;
```
JustinAWei, line 86: put this back

lib/translator/mode_sense.cc
harloff-g: Can't readval here because neither types are a span. We can provide more type safety with static_assert(sizeof(nvme_cmd) == sizeof(tmp_get_features))

lib/scsi.h
```diff
@@ -884,6 +903,144 @@ struct UnmapBlockDescriptor {
 } ABSL_ATTRIBUTE_PACKED;
 static_assert(sizeof(UnmapBlockDescriptor) == 16);
 
+// SCSI Reference Manual Table 372
+// https://www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf
+struct CachingModePage {
```
toadchild, line 908: Please double check the order of bit-fields in structs in this PR.

lib/translator/mode_sense.cc
toadchild: You could cast the pointer to a byte array and construct a span over that, using sizeof() to determine the length.

harloff-g: Does doing this actually provide any further safety or is it aesthetic?

lib/scsi.h
harloff-g: Great catch, there were a number of fields out of order

lib/translator/mode_sense.cc
Ralith: I agree that the benefit is dubious. Given the static assert I'd leave it as-is for simplicity.

lib/translator/mode_sense.cc
```diff
@@ -0,0 +1,402 @@
...
+
+#include <byteswap.h>
+
+namespace translator {
+
+namespace {
+
+// Section 6.3.3.1
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+constexpr scsi::CachingModePage caching_mode_page = {
```
Ralith, line 31:
```suggestion
constexpr scsi::CachingModePage kCachingModePage = {
```

lib/translator/mode_sense.cc
```diff
@@ -0,0 +1,402 @@
...
+    .page_code = scsi::ModePageCode::kCacheMode,
+    .spf = 0,
+    .ps = 0,
+    .page_length = 0x12,
+    .ic = 1,
+};
+
+// Section 6.3.3.2
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+constexpr scsi::ControlModePage control_mode_page = {
```
Ralith, line 41:
```suggestion
constexpr scsi::ControlModePage kControlModePage = {
```

lib/translator/mode_sense.cc
```diff
@@ -0,0 +1,402 @@
...
+    .rac = 0,
+    .autoload_mode = 0,
+    .tas = 0x01,
+    .ato = 0,
+    .busy_timeout_period = 0xFFFF,
+    .estct = 0};
+
+// Section 6.3.3.3
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+constexpr scsi::PowerConditionModePage power_condition_mode_page = {
```
Ralith, line 66:
```suggestion
constexpr scsi::PowerConditionModePage kPowerConditionModePage = {
```

lib/translator/mode_sense.h
```diff
@@ -0,0 +1,45 @@
...
+#ifndef LIB_TRANSLATOR_MODE_SENSE_H
+#define LIB_TRANSLATOR_MODE_SENSE_H
+
+#include "common.h"
+
+#include "third_party/spdk/nvme.h"
+
+namespace translator {
+
+StatusCode ModeSense6ToNvme(Span<const uint8_t> scsi_cmd,
```
eyhhuang, line 24: please add comment about what mode sense gets translated into

lib/translator/BUILD
JustinAWei: :common

lib/translator/mode_sense.h
harloff-g: similarly to #134 , links to translation specs are in the .cc, should we also place them in the .h?

harloff-g: Follow discussion there, a note will be made in the .h

lib/translator/mode_sense.cc
```diff
@@ -0,0 +1,403 @@
...
+
+// Writes corresponding mode page data to buffer
+void WritePageData(scsi::ModePageCode page_code, uint32_t get_features_result,
+                   Span<uint8_t> buffer) {
+  switch (page_code) {
+    case scsi::ModePageCode::kCacheMode: {
+      scsi::CachingModePage tmp_page = kCachingModePage;
+      // Figure 281 NVMe Spec
+      tmp_page.wce = 0b1 & get_features_result;
+      WriteValue(tmp_page, buffer);
```
eyhhuang, line 260: do you think it would be useful to have a check for these WriteValues

harloff-g: Good point, I updated the code to support this

PR #130, opened 2020-07-14
implement htonll for uint64_t

eyhhuang:
We currently use `htonl()` and `htons()` to ensure our SCSI structs are in big endian. However, there is currently no function for `uint64_t` which is the data type for some SCSI fields such as `logical_block_address`. 

`htonll()` first checks whether the machine is in big endian and then does the transformation if not.

### DIFF
--- a/lib/translator/common.cc
+++ b/lib/translator/common.cc
@@ -75,21 +75,6 @@ uint64_t htonll(uint64_t value) {
   }
 }
 
-uint64_t ntohll(uint64_t value) {
-  static uint32_t test_val = 42;
-
-  // Check first byte to determine endianness
-  if (*reinterpret_cast<const char*>(&test_val) == test_val) {  // little endian
-    const uint32_t high_bits = htonl(static_cast<uint32_t>(value >> 32));
-    const uint32_t low_bits =
-        htonl(static_cast<uint32_t>(value & 0xFFFFFFFFLL));
-
-    return (static_cast<uint64_t>(low_bits) << 32) | high_bits;
-  } else {  // big endian
-    return value;
-  }
-}
-
 StatusCode Allocation::SetPages(uint16_t data_page_count,
                                 uint16_t mdata_page_count) {
   if ((data_page_count != 0 && this->data_addr != 0) ||
--- a/lib/translator/common.cc
+++ b/lib/translator/common.cc
@@ -64,13 +64,28 @@ uint64_t htonll(uint64_t value) {
   static uint32_t test_val = 42;
 
   // Check first byte to determine endianness
-  if (*reinterpret_cast<const char*>(&test_val) == test_val) {
+  if (*reinterpret_cast<const char*>(&test_val) == test_val) {  // little endian
     const uint32_t high_bits = htonl(static_cast<uint32_t>(value >> 32));
     const uint32_t low_bits =
         htonl(static_cast<uint32_t>(value & 0xFFFFFFFFLL));
 
     return (static_cast<uint64_t>(low_bits) << 32) | high_bits;
-  } else {
+  } else {  // big endian
+    return value;
+  }
+}
+
+uint64_t ntohll(uint64_t value) {
+  static uint32_t test_val = 42;
+
+  // Check first byte to determine endianness
+  if (*reinterpret_cast<const char*>(&test_val) == test_val) {  // little endian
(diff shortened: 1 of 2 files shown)


lib/translator/common.cc
```diff
@@ -14,6 +14,7 @@
 
 #include "common.h"
 
+#include <netinet/in.h>
```
harloff-g, line 17: user-space includes are not available in the kernel

harloff-g:
We'll most likely have to handle endian-tranformations using callback functions. Interfaces for translating between big & little endian are notably different between user-space and kernel. In particular, htonl is not available in the kernel under the same includes as the kernel, and is facilitated by the kernel functions https://www.kernel.org/doc/htmldocs/kernel-hacking/routines-endian.html

lib/translator/common.cc
```diff
@@ -54,6 +60,21 @@ void SetAllocPageCallbacks(uint64_t (*alloc_callback)(uint16_t),
   dealloc_pages_callback = dealloc_callback;
 }
 
+uint64_t htonll(uint64_t value) {
+  static int test_val = 42;
```
harloff-g, line 64: Can we make this a uint32_t to remove ambiguity from bit-sensitive logic

lib/translator/common.cc
```diff
@@ -54,6 +60,21 @@ void SetAllocPageCallbacks(uint64_t (*alloc_callback)(uint16_t),
...
+
+  // Check first byte to determine endianness
+  if (*reinterpret_cast<const char*>(&test_val) == test_val) {
```
harloff-g, line 67: Can we add a comment documenting which branch corresponds to which endianness

harloff-g:
In general, does this meet the requirements of our library? In particular, is the "endianness translation direction" of htonl well defined (i.e. htonl always returns a big endian)? Our library needs to be able to explicitly convert to big endian and little endian without regard to network byte order, as these are the endian layouts required respectively by the SCSI and NVMe spec.

eyhhuang:
> In general, does this meet the requirements of our library? In particular, is the "endianness translation direction" of htonl well defined (i.e. htonl always returns a big endian)? Our library needs to be able to explicitly convert to big endian and little endian without regard to network byte order, as these are the endian layouts required respectively by the SCSI and NVMe spec.

The network byte order is defined to always be big endian, so htonl(), htons(), and htonll() will always result in big endian. The 
"network byte order" is defined by the IETF and isn't dependent on any of our internal configuations.
 These functions will be used ensure the SCSI commands are laid out in big endian. Since SCSI and NMVe have opposite endianness, and after using htonl() we know that SCSI will be in big endian, we can use byteswap to ensure the NVMe commands are laid out in little endian. 

I will add some comments where this function is defined to clarify its function and intended usage.

eyhhuang:
> Do we also plan on implementing ntohll?

I'm actually not sure when we would need the ntohl functions, since the behaviour would depend based on the host.

lib/translator/common.cc
```diff
@@ -54,6 +60,36 @@ void SetAllocPageCallbacks(uint64_t (*alloc_callback)(uint16_t),
...
+  } else {  // big endian
+    return value;
+  }
+}
+
+uint64_t ntohll(uint64_t value) {
+  static uint32_t test_val = 42;
+
+  // Check first byte to determine endianness
+  if (*reinterpret_cast<const char*>(&test_val) == test_val) {  // little endian
```
harloff-g, line 82: If these two functions are the exact same internally should we consider using `"#define ntohll htonll"` for reduced code duplication? I think having both names even if they do the same thing does make sense from a self-documenting standpoint

eyhhuang: I think using #define is a good idea!

lib/translator/common.cc
```diff
@@ -54,6 +60,21 @@ void SetAllocPageCallbacks(uint64_t (*alloc_callback)(uint16_t),
   dealloc_pages_callback = dealloc_callback;
 }
 
+uint64_t htonll(uint64_t value) {
+  static uint32_t test_val = 42;
+
+  // Check first byte to determine endianness
+  if (*reinterpret_cast<const char*>(&test_val) == test_val) {  // little endian
```
toadchild, line 67: There should be some standard compiler-provded preprocessor macros you can check so that you can use `#ifdef` instead of `if(...)`, but this is probably sufficient for our purposes at this time.

PR #134, opened 2020-07-16
Translate Unmap

### LINKED ISSUE #129: Translate Unmap Command

harloff-g:
Closes #129 

Also adds a "buffer_out" field to the Begin function. This is because variable-length command data is stored in a separate buffer than the fixed-length command in SCSI.


lib/translator/BUILD
```diff
@@ -65,3 +66,13 @@ cc_library(
   ],
   visibility = ["//visibility:public"],
 )
+
+cc_library(
+  name = "unmap_lib",
+  hdrs = ["unmap.h"],
+  srcs = ["unmap.cc"],
+  deps = [
+      "//lib/translator:common"
```
JustinAWei, line 75: just :common

lib/translator/unmap.cc
```diff
@@ -0,0 +1,92 @@
...
+
+namespace translator {
+
+// Section 5.6
+// https://www.nvmexpress.org/wp-content/uploads/NVM-Express-SCSI-Translation-Reference-1_1-Gold.pdf
+StatusCode UnmapToNvme(absl::Span<const uint8_t> scsi_cmd,
+                       absl::Span<const uint8_t> buffer_out,
+                       nvme::GenericQueueEntryCmd& nvme_cmd,
+                       Allocation& allocation, uint32_t nsid) {
+  scsi::UnmapCommand unmap_cmd;
```
JustinAWei, line 25: = {} for zero

lib/translator/unmap.cc
```diff
@@ -0,0 +1,92 @@
...
+    DebugLog("Insufficient unmap parameter list length");
+    return StatusCode::kFailure;
+  }
+  if (unmap_cmd.anchor == 1) {
+    DebugLog("Unsupported unmap anchor request");
+    return StatusCode::kNoTranslation;
+  }
+
+  // Copy unmap parameter list into memory
+  scsi::UnmapParamList param_list;
```
JustinAWei, line 40: {}

lib/translator/unmap.cc
```diff
@@ -0,0 +1,92 @@
...
+  }
+  buffer_out = buffer_out.subspan(sizeof(scsi::UnmapParamList));
+
+  // Ensure that block descriptor data length & span length align
+  if (buffer_out.size() < param_list.block_desc_data_length) {
+    DebugLog("Block descriptor list length reported longer than buffer");
+    return StatusCode::kFailure;
+  }
+  uint32_t block_descriptor_count =
+      param_list.block_desc_data_length / sizeof(scsi::UnmapBlockDescriptor);
```
JustinAWei, line 53: do we need to worry about divisibility?

lib/translator/unmap.cc
```diff
@@ -0,0 +1,92 @@
...
+  if (block_descriptor_count == 0 || block_descriptor_count > 256) {
+    DebugLog("Unsupported unmap block descriptor count %u",
+             block_descriptor_count);
+    return StatusCode::kNoTranslation;
+  }
+
+  // Copy block descriptors into structures
+  scsi::UnmapBlockDescriptor block_descriptors[block_descriptor_count];
```
JustinAWei, line 61: = {{}}

lib/translator/unmap.cc
```diff
@@ -0,0 +1,92 @@
...
+      return StatusCode::kFailure;
+    }
+    buffer_out = buffer_out.subspan(sizeof(block_descriptors[i]));
+  }
+
+  // Copy block descriptor data into allocated nvme data buffer
+  if (allocation.SetPages(1, 0) == StatusCode::kFailure)
+    return StatusCode::kFailure;
+  nvme::DatasetManagmentRange* dmr_ptr =
+      reinterpret_cast<nvme::DatasetManagmentRange*>(allocation.data_addr);
```
JustinAWei, line 74: use ReadValue in this case still for bounds checkin

lib/translator/unmap.cc
```diff
@@ -0,0 +1,92 @@
...
+  for (uint32_t i = 0; i < block_descriptor_count; ++i) {
+    dmr_ptr[i].lba = block_descriptors[i].logical_block_addr;
+    dmr_ptr[i].lb_count = block_descriptors[i].logical_block_count;
+  }
+
+  // Create NVMe command
+  uint8_t bd_count_byte = static_cast<uint8_t>(block_descriptor_count - 1);
+  nvme::DatasetManagementCmd dataset_cmd = {
+      .opc = static_cast<uint8_t>(nvme::NvmOpcode::kDatasetManagement),
```
JustinAWei, line 83: why not make opc NvmOpcode in the struct def

lib/translator/unmap.cc
```diff
@@ -0,0 +1,92 @@
...
+      .nsid = nsid,
+      .nr = bd_count_byte,  // (1's based -> 0's based)
+      .ad = 1};
+  dataset_cmd.dptr.prp.prp1 = allocation.data_addr,
+  memcpy(&nvme_cmd, &dataset_cmd, sizeof(nvme_cmd));
```
JustinAWei, line 88: readval

lib/translator/unmap.cc
harloff-g: Why is it necessary to zero memory that will be immediately overwritten by readvalue?

lib/translator/unmap.cc
harloff-g: Other PRs in master have already used .opc = static_cast<uint8_t>(nvme::NvmOpcode::kDatasetManagement) so such a change would be dependent on a dedicated PR

lib/translator/unmap.cc
harloff-g: Good catch, technically a non-divisible block_desc_data_length is a SCSI check condition error. I'll update accordingly

lib/translator/unmap.cc
harloff-g: Resolving per recent kernel stack discoveries

lib/translator/translation.cc
```diff
@@ -87,15 +96,18 @@ ApiStatus Translation::Complete(Span<const nvme::GenericQueueEntryCpl> cpl_data,
 
   switch (opc) {
     case scsi::OpCode::kInquiry:
-      pipeline_status_ = InquiryToScsi(scsi_cmd_no_op, buffer, GetNvmeCmds());
+      pipeline_status_ =
+          InquiryToScsi(scsi_cmd_no_op, buffer_in, GetNvmeCmds());
       ret = ApiStatus::kSuccess;
       break;
+    case scsi::OpCode::kUnmap:
+      break;
```
toadchild, line 104: Do you not need to call a function here to set pipeline_status_ and check for errors from nvme?

lib/translator/unmap.cc
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  // Copy block descriptor data into allocated nvme data buffer
+  if (allocation.SetPages(1, 0) == StatusCode::kFailure)
+    return StatusCode::kFailure;
+  nvme::DatasetManagmentRange* dmr_ptr =
+      reinterpret_cast<nvme::DatasetManagmentRange*>(allocation.data_addr);
+  for (uint32_t i = 0; i < block_descriptor_count; ++i) {
+    dmr_ptr[i].lba = block_descriptors[i].logical_block_addr;
+    dmr_ptr[i].lb_count = block_descriptors[i].logical_block_count;
```
toadchild, line 84: These values need to be byte-swapped.

lib/translator/translation.cc
harloff-g: #150 handles errors independent of command logic. As of now pipeline_status_ doesn't do anything after this point in the logic (is set to invalid by AbortPipeline at the end of complete), but given all other cases have taken to setting it I'll add a setter here as well. No function is needed in general for unmap because there is no response data passed to the SCSI buffer.

lib/translator/unmap.cc
harloff-g: Would like to hear others opinions on the benefits of creating a span of size nvme_cmd just to call readval with. Seems superficial to me

Ralith: Given the `static_assert` below I think this is fine as-is. Those helpers are mostly for when we *don't* statically know the size and type of the source data.

test/translator/translation_test.cc
```diff
@@ -26,15 +26,15 @@ TEST(Translation, ShouldHandleUnknownOpcode) {
   translator::Translation translation = {};
   uint8_t opc = 233;
   translator::Span<const uint8_t> scsi_cmd = translator::Span(&opc, 1);
-  translator::BeginResponse resp = translation.Begin(scsi_cmd, 0);
+  translator::BeginResponse resp = translation.Begin(scsi_cmd, scsi_cmd, 0);
```
Ralith, line 29: Is it sensible to pass in the same buffer for both arguments here? If this case isn't expected to read from the second buffer, then a default-constructed empty span might make more sense.

lib/translator/common.h
```diff
@@ -16,6 +16,7 @@
 #define LIB_TRANSLATOR_COMMON_H
 
 #include <cstring>
+#include <new>
```
eyhhuang, line 19: I think this can just be placed in unmap.cc

lib/translator/unmap.h
```diff
@@ -0,0 +1,30 @@
...
+#ifndef LIB_TRANSLATOR_UNMAP_H
+#define LIB_TRANSLATOR_UNMAP_H
+
+#include "common.h"
+
+#include "third_party/spdk/nvme.h"
+
+namespace translator {
+
+StatusCode UnmapToNvme(Span<const uint8_t> scsi_cmd,
```
eyhhuang, line 24: can we add a comment here describing what nvme command unmap corresponds to and maybe also describing what buffer_out is?

test/translator/unmap_test.cc
```diff
@@ -0,0 +1,92 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "lib/translator/unmap.h"
+
+#include "gtest/gtest.h"
+
+#include <byteswap.h>
```
eyhhuang, line 19: nit: did a quick ctrl f and it looks like byteswap isn't used in this test

lib/translator/unmap.cc
JustinAWei: oh, you're right

lib/translator/common.h
harloff-g: It's required here because it's used by SafePointerCastWrite. Should have been in the original PR but for some reason tests didn't catch it

lib/translator/unmap.h
harloff-g: I'll go ahead and add a comment about buffer_out. The unmap NVMe equivalent is linked in the .cc file. Should we put it here as well?

test/translator/unmap_test.cc
harloff-g: nice catch

lib/translator/unmap.h
eyhhuang: Oh I see the comment in the .cc. I think it'd be helpful in the header file to say that Unmap translates to an NVMe Dataset Management command

lib/translator/translation.cc
Ralith: Was this addressed? I still don't see an assignment to pipeline_status_.

harloff-g: We can set pipeline_status_ to success to be more explicit, but any errors stemming from the unmap command are handled upstream of this logic

Ralith: Makes sense, but we indeed should try to be locally consistent as well.

PR #135, opened 2020-07-17
Translate read response

eyhhuang:
Closes https://github.com/googleinterns/scsi2nvme/issues/93


harloff-g:
Regarding endianness; I'm not positive but I think that since the read data is just the raw data on device, which is "communication protocol independent" in some sense, we don't have to account for endianness

toadchild:
Re: endianness - raw data on the device should be passed through with no alterations.

lib/translator/read.cc
```diff
@@ -198,4 +198,13 @@ StatusCode Read16ToNvme(absl::Span<const uint8_t> scsi_cmd,
               nvme_cmd, allocation, nsid);
 }
 
+void ReadToScsi(absl::Span<uint8_t> buffer,
+                const nvme::GenericQueueEntryCmd& nvme_cmd) {
+  uint64_t data_ptr = nvme_cmd.dptr.prp.prp1;
```
toadchild, line 203: Do we have any functions to parse prp lists if the data is bigger than a single page?

lib/translator/read.cc
```diff
@@ -194,4 +194,13 @@ StatusCode Read16ToNvme(Span<const uint8_t> scsi_cmd,
               nvme_cmd, allocation, nsid);
 }
 
+void ReadToScsi(Span<uint8_t> buffer,
+                const nvme::GenericQueueEntryCmd& nvme_cmd) {
+  uint64_t data_ptr = nvme_cmd.dptr.prp.prp1;
+  // cdw 12 nlb bits 15:00 (zero based field)
+  uint16_t transfer_length = nvme_cmd.cdw[2] + 1;
+
+  memcpy(buffer.data(), reinterpret_cast<uint8_t*>(data_ptr), transfer_length);
```
JustinAWei, line 203: readval

eyhhuang: Can't really use ReadVal here because I need to specifically copy `transfer_length` number of bytes :(

lib/translator/read.cc
eyhhuang: Discussed this offline but it seems like the NVMe driver is responsible for managing the prp list if there are multiple pages of data. This is supported by how the nvme cli handles the buffer https://github.com/linux-nvme/nvme-cli/blob/master/nvme.c#L4094. All that we need to do on our end is to allocate a sufficiently large buffer, set the prp pointer to the pointer to the buffer, and then the buffer will have the response data copied into it. So when translating back to SCSI, we can just copy the data pointed to by the prp to the scsi data in buffer.

PR #140, opened 2020-07-20
Remove Problematic STL Includes

### LINKED ISSUE #139: Remove all dependencies on STL in codebase
Almost all STL dependencies do not play nice with kernel. Removing these dependencies will make it possible for the library to run in a kernel module.

harloff-g:
Closes #139 

Our library currently uses a number of includes that are not compatible with use in the Linux kernel. In particular, absl::Span has a great deal of dependencies on the STL.

This PR set out to remove all of these dependencies. Most notably, absl::Span was abandoned because of the large refactor time it would take to make it work without kernel-breaking stl features. A new Span class was written from the ground up to support our current use case.

### DIFF
--- a/lib/translator/common.h
+++ b/lib/translator/common.h
@@ -18,9 +18,6 @@
 #include <cstring>
 #include <type_traits>
 
-#include "absl/strings/string_view.h"
-#include "absl/types/span.h"
-
 #include "lib/scsi.h"
 #include "third_party/spdk/nvme.h"
 
@@ -28,8 +25,8 @@ namespace translator {
 
 // Vendor Identification shall be set to "NVMe" followed by 4 spaces: "NVMe    "
 // This value is not null terminated and should be size 8
-constexpr absl::string_view kNvmeVendorIdentification = "NVMe    ";
-static_assert(kNvmeVendorIdentification.size() == 8);
+constexpr char kNvmeVendorIdentification[] = "NVMe    ";
+static_assert(strlen(kNvmeVendorIdentification) == 8);
 
 // The maximum amplification ratio of any supported SCSI:NVMe translation
 constexpr int kMaxCommandRatio = 3;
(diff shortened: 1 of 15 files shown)


lib/translator/common.h
```diff
@@ -70,15 +67,41 @@ void SetAllocPageCallbacks(uint64_t (*alloc_callback)(uint16_t),
...
-bool ReadValue(absl::Span<const uint8_t> data, T& out) {
+class Span {
+ public:
+  Span() : ptr_(nullptr), len_(0) {}
+  Span(T* ptr, size_t len) : ptr_(ptr), len_(len) {}
+  template <size_t N>
+  Span(T (&a)[N]) : Span(a, N) {}
+  template <typename Y,
+            typename = std::enable_if<std::is_convertible<T*, Y*>::value>>
+  Span(const Span<Y>& ref) : Span(ref.data(), ref.size()) {}
```
Ralith, line 78: I don't think this is sound, as a derived class pointer is convertible to a base class but may have different size, in which case pointer arithmetic will give incorrect results.

Do we actually need this interface at all?

harloff-g: Good point, being more explicit here is best. We need this interface because the compiler doesn't implicitly know how to copy construct from Span<T> => Span<const T> (at least on my machine). This is something that ABSL supports and our library uses. I've pushed an update to the enable_if condition which should alleviate concerns of type size differences, and still enable the intended use case

harloff-g:
> What were the STL interfaces that absl::Span used which broke the build?

I noticed build errors stemming from:
#include <algorithm>
#include <iterator>
#include <utility>
#include <functional>
#include <new>
#include <stdexcept>

Ralith:
Oh, right, the headers don't work because those end up in the kernel build, which excludes standard include paths. Guess there's nothing for it.

lib/translator/common.h
Ralith: > We need this interface because the compiler doesn't implicitly know how to copy construct from Span<T> => Span<const T>

Ah, of course.

> I've pushed an update to the enable_if condition which should alleviate concerns of type size differences, and still enable the intended use case

That looks better, thanks. It might be safer still to more closely duplicate [what abseil does](https://github.com/abseil/abseil-cpp/blob/12bc53e0318d80569270a5b26ccbc62b52022b89/absl/types/internal/span.h#L59-L68) here, but your approach is simpler and there's a lot to be said for that, in the absence of any known shortcomings.

Ralith:
I'm a bit confused (albeit relieved) by the implication that `#include <type_traits>` works when all those others don't. Out of curiosity, have you found a consistent rule for what headers are and aren't available?

harloff-g:
> I'm a bit confused (albeit relieved) by the implication that `#include <type_traits>` works when all those others don't. Out of curiosity, have you found a consistent rule for what headers are and aren't available?

I've noticed that most STL includes fail due to not complying with flags set in the kernel compilation unit intended for memory safety/C compatibility.

In particular, the flags "-fno-rtti" and "-fno-exceptions" cause many stl includes to fail

Ralith:
> In particular, the flags "-fno-rtti" and "-fno-exceptions" cause many stl includes to fail

Can you elaborate on that a bit? AFAIK STL is routinely used with those flags; [a quick test](https://godbolt.org/z/x8E7b7) shows `std::string` working fine with both enabled.

I figured the includes just weren't being found at all due to how the kernel build system's set up. What specific errors did you see?

harloff-g:
> > In particular, the flags "-fno-rtti" and "-fno-exceptions" cause many stl includes to fail
> 
> Can you elaborate on that a bit? AFAIK STL is routinely used with those flags; [a quick test](https://godbolt.org/z/x8E7b7) shows `std::string` working fine with both enabled.
> 
> I figured the includes just weren't being found at all due to how the kernel build system's set up. What specific errors did you see?

Sorry, looks like I got some flags confused in the build log. The errors are mostly of two forms:
`Error: x86 register return with SSE disabled`
and
`Error: SSE register return with SSE disabled`

Ralith:
Interesting, thanks for the details! Looks like floating point is simply Not Done in the kernel.

Issue #138, opened 2020-07-20
Correct bit order in SCSI


eyhhuang:
Bit fields in SCSI should be arranged from smallest to largest. NVMe structs have their bit fields correctly arranged

PR #141, opened 2020-07-20
Order scsi structs

eyhhuang:
This change ensures that all SCSI structs have bit fields laid out from smallest bit to largest.

While making this change, I noticed that some SCSI structs followed the SPC-3 specs instead of SPC-4, so this change makes sure all structs follow SPC-4. Changing from SPC-3 to SPC-4 mainly involves replacing fields such as `fua_nv` with obsolete or changing the size of fields such as `group_number`. The fields affected are not involved in the translation to NVMe structs. 

Finally, I removed the Read/Write(32) structs and corrected some typos and incorrect table numbers in comments

Question: Should we somehow test that these structs are correctly laid out for extra assurance? 

Closes https://github.com/googleinterns/scsi2nvme/issues/138


lib/translator/inquiry.cc
```diff
@@ -171,9 +171,9 @@ void TranslateBlockLimitsVpd(const nvme::IdentifyControllerData& identify_ctrl,
   // TODO: put in common?
   const uint8_t kMaxCompareWriteLen = 255;
   // compare and write length is 8 bits. we ensure the size fits.
-  uint32_t compare_and_write_len = (max_transfer_length > kMaxCompareWriteLen)
-                                       ? kMaxCompareWriteLen
-                                       : max_transfer_length;
+  uint8_t compare_and_write_len = (max_transfer_length > kMaxCompareWriteLen)
```
eyhhuang, line 174: As per the comment in line 173, this field is supposed to be 8 bits. I made the change to get rid of a compiler narrow conversion warning

lib/scsi.h
```diff
@@ -416,67 +414,44 @@ static_assert(sizeof(Read10Command) == 9);
...
-  bool fua_nv : 1;  // Forced Access Unit Non-Volatile
-  bool reserved1 : 1;
   uint64_t logical_block_address : 64;
   uint32_t transfer_length : 32;
-  bool restricted_mmc_6 : 1;
-  uint8_t reserved2 : 2;
-  uint8_t group_number : 5;
+  uint8_t group_number : 6;
+  bool dld_0 : 1;
+  bool dld_1 : 1;
```
toadchild, line 444: This section has a bigger than normal change; group_number changed size, and the mmc field was removed entirely.  This is in addition to the reserved field being replaced with dld.

eyhhuang: Yeah, while I was making the changes just for the bit field ordering, I realized that some of the structs followed the SPC-3 specs instead of SPC-4 so I made them all consistent with SPC-4

toadchild: Sounds good, thanks for clarifying.

PR #142, opened 2020-07-21
Maintenance in

eyhhuang:
For the MVP, the only `MaintenanceIn` command we support is the `ReportSupportOpCodes` command. For `ReportSupportedOpCodes`, we only support whether `WriteSame16` is supported. Currently we always report that `WriteSame16` is not supported.

This operation does not require a call to the NVMe device. So, instead of having a `ToNvme` and a `ToScsi` function, it has `Validate` and `WriteResponse`, which are called in `Translation::Begin` and `Translation::End` respectively

Closes https://github.com/googleinterns/scsi2nvme/issues/133


lib/translator/maintenance_in.cc
```diff
@@ -0,0 +1,47 @@
...
+  scsi::ReportOpCodesCommand report_cmd = {};
+  if (!ReadValue(scsi_cmd, report_cmd)) {
+    DebugLog("Malformed Report Supported OpCodes command");
+    return StatusCode::kInvalidInput;
+  }
+
+  if (report_cmd.requested_op_code !=
+          static_cast<uint8_t>(scsi::OpCode::kWriteSame16) ||
+      report_cmd.reporting_options != 0b001) {
+    DebugLog("Only supporting ReportSupportedOpCodes for WriteSame16");
```
harloff-g, line 31: Why do we only support this one niche case?

eyhhuang: This was in the project specs: " only report if we support “WriteSame16”. I think it might be because on start up/connection the scsi device queries whether WriteSame16 is supported?

JustinAWei: good to add a comment about this

eyhhuang: there's a comment in the header file describing this but it definitely makes a lot of sense to put a comment here

PR #150, opened 2020-07-22
Handle Sense Data and Errors

harloff-g:
1. Fixes bit fields in SCSI sense data structures.
2. Adds helper function to populate a sense buffer
3. Adds sense buffer as parameter to Translation::Complete
4. Adds logic to Translation::Complete which calls sense_fill_buffer and returns status codes to the API consumer which will be needed for upstream function returns.


lib/translator/common.h
```diff
@@ -120,5 +120,16 @@ bool WriteValue(const T& data, Span<uint8_t> out) {
...
+// Scsi status bundle
+struct ScsiStatus {
+  scsi::Status status;
+  scsi::SenseKey sense_key;
+  scsi::AdditionalSenseCode asc;
+  scsi::AdditionalSenseCodeQualifier ascq;
+};
+
+// Fills sense buffer with relevant data. Returns true on success.
+bool fill_sense_buffer(Span<uint8_t> sense_buffer, ScsiStatus& scsi_status);
```
eyhhuang, line 132: nit: I think this should be FillSenseBuffer

lib/translator/translation.cc
```diff
@@ -68,39 +69,68 @@ BeginResponse Translation::Begin(Span<const uint8_t> scsi_cmd,
   return response;
 }
 
-ApiStatus Translation::Complete(Span<const nvme::GenericQueueEntryCpl> cpl_data,
-                                Span<uint8_t> buffer) {
+CompleteResponse Translation::Complete(
+    Span<const nvme::GenericQueueEntryCpl> cpl_data, Span<uint8_t> buffer,
```
eyhhuang, line 73: nit: might want to rename to `buffer_in` so it's consistent with the name in the header

test/translator/translation_test.cc
```diff
@@ -42,8 +42,10 @@ TEST(Translation, ShouldFailInvalidPipeline) {
   translator::Translation translation = {};
   translator::Span<const nvme::GenericQueueEntryCpl> cpl_data;
   translator::Span<uint8_t> buffer;
```
eyhhuang, line 44:
```suggestion
  translator::Span<uint8_t> buffer_in;
```

test/translator/translation_test.cc
```diff
@@ -42,8 +42,10 @@ TEST(Translation, ShouldFailInvalidPipeline) {
...
-  translator::ApiStatus status = translation.Complete(cpl_data, buffer);
-  EXPECT_EQ(translator::ApiStatus::kFailure, status);
+  translator::Span<uint8_t> sense_buffer;
+  translator::CompleteResponse status =
+      translation.Complete(cpl_data, buffer, sense_buffer);
```
eyhhuang, line 47:
```suggestion
      translation.Complete(cpl_data, buffer_in, sense_buffer);
```

lib/translator/translation.cc
harloff-g: Nice catch

lib/translator/common.cc
```diff
@@ -199,4 +199,23 @@ const char* ScsiOpcodeToString(scsi::OpCode opcode) {
   }
 }
 
+bool FillSenseBuffer(Span<uint8_t> sense_buffer, ScsiStatus& scsi_status) {
```
Ralith, line 202:
```suggestion
bool FillSenseBuffer(Span<uint8_t> sense_buffer, const ScsiStatus& scsi_status) {
```
Use const references whenever possible. This makes it clearer what the intention of the function is and allows it to be used in more contexts.

lib/translator/translation.h
```diff
@@ -29,6 +29,11 @@ struct BeginResponse {
   uint32_t alloc_len;  // Defines size of buffer passed to Translation::Complete
 };
 
+struct CompleteResponse {
+  ApiStatus status;
+  scsi::Status scsi_status;  // Return value of library consumer functions
```
Ralith, line 34: Why does the caller need access to this?

lib/translator/common.cc
```diff
@@ -199,4 +199,23 @@ const char* ScsiOpcodeToString(scsi::OpCode opcode) {
...
+  // Descriptor sense logic (d_sense set to 1)
+  // Table 12 Seagate SCSI Spec
+  scsi::DescriptorFormatSenseData dfsd;
+  if (!ReadValue(sense_buffer, dfsd)) {
```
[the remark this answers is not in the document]
harloff-g: We currently don't return additional sense data so I'll remove the readvalue and set additional_sense_length to 0. Originally this function was meant to supported both fixed & descriptor sense data, but after reading the SCSI to NVMe spec only descriptor format is supported

lib/translator/translation.h
harloff-g: The QueueCommand SCSI function hook expects a status code as a return value. The consumer needs to have access to this value to return it from the calling function

lib/translator/translation.cc
```diff
@@ -69,39 +70,69 @@ BeginResponse Translation::Begin(Span<const uint8_t> scsi_cmd,
...
+      FillSenseBuffer(sense_buffer, scsi_status);
+      resp.status = ApiStatus::kSuccess;
+      resp.scsi_status = scsi_status.status;
+      return resp;
+    }
   }
 
   // Switch cases should not return
-  ApiStatus ret;
+  resp.status = ApiStatus::kSuccess;
```
JustinAWei, line 119: on default switch case we are successful, is this correct behavior?

harloff-g: Technically we should never have a default switch case in this section of the library because the switch statement is validated in begin, which is enforced to be called before complete. Any default cases would imply our library is missing a switch that is in begin. We can log an error in this case

lib/translator/translation.cc
```diff
@@ -69,39 +70,73 @@ BeginResponse Translation::Begin(Span<const uint8_t> scsi_cmd,
...
-      ret = ApiStatus::kSuccess;
+      pipeline_status_ = ReadCapacity10ToScsi(buffer_in, nvme_cmds_[0]);
       break;
     case scsi::OpCode::kRequestSense:
-      pipeline_status_ = RequestSenseToScsi(scsi_cmd_no_op, buffer);
-      ret = ApiStatus::kSuccess;
+      pipeline_status_ = RequestSenseToScsi(scsi_cmd_no_op, buffer_in);
+      break;
+    default:
+      DebugLog("Invalid case reached. Please contact SCSI2NVMe devs.");
```
Ralith, line 135: Would it make sense to log the value of the unexpected opcode?

harloff-g: Good idea, I think that would be useful if this error ever occurs.

PR #156, opened 2020-07-24
Safe Pointer Casting Helper

### LINKED ISSUE #153: Add safe pointer casting
We can add helpers similar to Readvalue and Writevalue, but construct pointers rather than values. There are two additional subtleties that need to be accounted for in doing so:
- The memory must be suitably aligned. We might be able to guarantee this statically by documenting requirements for memory passed in by the caller and being careful with how we handle it internally. A safer approach, which is not mutually exclusive, would be to check alignment as well as size inside the helper functions themselves; we should probably start there, and then consider disabling it in release builds if we can document sufficient requirements that the check should never fail.
- Strict aliasing forbids treating a uint8* as a T* except under specific circumstances. When reading values, we can declare it the caller's responsibility to ensure the passed-in memory in question has the correct type in terms of the abstract language semantics. When writing values, we must use "placement new" to legally construct an object on top of existing untyped memory.

harloff-g:
Closes #153


lib/translator/common.h
```diff
@@ -123,5 +123,22 @@ bool WriteValue(const T& data, Span<uint8_t> out) {
...
+    return nullptr;
+  }
+  if (reinterpret_cast<uintptr_t>(buf.data()) % std::alignment_of<T>::value !=
+      0) {
+    DebugLog("Pointer cast called on unaligned memory");
+    return nullptr;
+  }
+  return new (buf.data()) T;
+}
+
```
Ralith, line 142: This is suitable for writing, but not for reading existing data: the contents are default-constructed, which for most of our purposes means undefined. We also need a version that does `Span<const uint8_t>` -> `const T*` which does not use placement new.

lib/translator/common.h
```diff
@@ -123,5 +123,22 @@ bool WriteValue(const T& data, Span<uint8_t> out) {
   return true;
 }
 
+// Returns a pointer of type T pointing to buf.data()
+// Uses placement new operator to garentee memory safety
+// Validates alignment and size of buffer
```
Ralith, line 128: This should state that it creates a new `T`; trying to access an existing one would be undefined behavior. Typo in "guarantee", also.

test/translator/common_test.cc
```diff
@@ -150,4 +150,20 @@ TEST(Common, ShouldFailBuildAllocationWhenAllocPageFails) {
...
+  else
+    expected_val = 0x22202913;
+  uint8_t buf[100];
+  buf[0] = 0x22;
+  buf[1] = 0x20;
+  buf[2] = 0x29;
+  buf[3] = 0x13;
+
+  uint32_t* val = translator::SafePointerCast<uint32_t>(buf);
+  ASSERT_EQ(expected_val, *val);
```
Ralith, line 166: Per above, this is undefined behavior.

lib/translator/common.h
harloff-g: Rather than using placement new for `const T*`, what method should we use to ensure the memory is not default-constructed?

Ralith: A plain `reinterpret_cast` is appropriate on the reading side. The principle is that we require the caller to have made the necessary arrangements for that to be sound, so we can just cast.

test/translator/common_test.cc
```diff
@@ -150,4 +150,37 @@ TEST(Common, ShouldFailBuildAllocationWhenAllocPageFails) {
...
+  else
+    expected_val = 0x22202913;
+  uint8_t buf[100];
+  buf[0] = 0x22;
+  buf[1] = 0x20;
+  buf[2] = 0x29;
+  buf[3] = 0x13;
+
+  const uint32_t* val = translator::SafePointerCastRead<uint32_t>(buf);
+  ASSERT_EQ(expected_val, *val);
```
Ralith, line 183: This is undefined behavior because it treats a `uint8_t` array as a `uint32_t`, violating strict aliasing. A legal approach would be to instead declare a `uint32_t` to begin with, then use `reinterpret_cast` to convert that into a `uint8_t*` to read from.

test/translator/common_test.cc
```diff
@@ -151,35 +151,25 @@ TEST(Common, ShouldFailBuildAllocationWhenAllocPageFails) {
...
-
-  uint32_t* val = translator::SafePointerCastWrite<uint32_t>(buf);
+  uint32_t expected_val = 0x13292022;
+
+  uint32_t write_buffer = 0;
+  uint8_t* buf = reinterpret_cast<uint8_t*>(&write_buffer);
+  translator::Span<uint8_t> span(buf, sizeof(uint32_t));
+
+  uint32_t* val = translator::SafePointerCastWrite<uint32_t>(span);
   *val = expected_val;
```
Ralith, line 161: Both tests should check that the returned pointer isn't `nullptr` before dereferencing it.

test/translator/common_test.cc
```diff
@@ -150,4 +150,27 @@ TEST(Common, ShouldFailBuildAllocationWhenAllocPageFails) {
...
+  uint32_t expected_val = 0x13292022;
+  uint8_t* buf = reinterpret_cast<uint8_t*>(&expected_val);
+  translator::Span<uint8_t> span(buf, sizeof(uint32_t));
+
+  const uint32_t* val = translator::SafePointerCastRead<uint32_t>(span);
+
+  ASSERT_EQ(expected_val, *val);
+}
+
 }  // namespace
```
toadchild, line 176: Do you want tests that cover error cases?

harloff-g: Good point, I've gone ahead and added tests which cover all error cases

PR #158, opened 2020-07-28
Enabling End to End Translation

basimsahaf:
The aim of this PR is to add the main engine code which runs the entire translation. The engine is responsible for initializing the SCSI mock, NVMe device and translation engine. In addition to this, the engine acts as an anchor between all these components. To get more context on the design of the project, refer to the internal design document which outlines each component in detail.

Items to review:
- SCSI Mock Module code in `scsi_mock_module.c`
- NVMe device communication module in `nvme_driver.c`
- Engine code in `engine.cc` and `engine.h`

There might be some other changes made in this PR while performing the e2e testing. 

Note: This PR was formed after merging a few other PRs and this is the reason that this PR is so big. In addition to that, all this code was required at once since we need to test the  engine with all this code. It is also easier for the reviewers to view the code at once since there are multiple moving pieces.


third_party/e2e/engine.cc
```diff
@@ -0,0 +1,90 @@
+#include "engine.h"
```
oicirtap, line 1: Is there a specific reason why these files and in general the e2e directory is under third_party? Is it just that this is not part of the library?

lib/translator/common.h
```diff
@@ -27,8 +27,7 @@ namespace translator {
 
 // Vendor Identification shall be set to "NVMe" followed by 4 spaces: "NVMe    "
 // This value is not null terminated and should be size 8
-constexpr char kNvmeVendorIdentification[] = "NVMe    ";
-static_assert(strlen(kNvmeVendorIdentification) == 8);
```
oicirtap, line 31: Is it possible to maintain this checks after swapping to using char*? The goal here is to prevent people from changing the length of the constant by mistake.

third_party/e2e/engine.cc
```diff
@@ -0,0 +1,90 @@
...
+#include "lib/translator/translation.h"
+#include "nvme_driver.h"
+
+namespace {
+
+uint64_t AllocPagesCallback(uint16_t count) { return AllocPages(count); }
+
+void DeallocPagesCallback(uint64_t addr, uint16_t count) {
+  DeallocPages(addr, count);
+}
```
oicirtap, line 14: Any reason not to use the underlying functions as callbacks directly?

third_party/e2e/engine.cc
```diff
@@ -0,0 +1,90 @@
...
+  translator::Translation translation;
+
+  // Package parameters and run translation begin
+  translator::Span<uint8_t> scsi_cmd(cmd_buf, cmd_len);
+  translator::Span<uint8_t> buffer(data_buf, data_len);
+  translator::BeginResponse begin_resp =
+      translation.Begin(scsi_cmd, buffer, lun);
+
+  if (begin_resp.status == translator::ApiStatus::kFailure) {
+    Print("Incorrect usage of Translation Library API");
```
oicirtap, line 37: Is this guaranteed to be an incorrect usage from part of the caller? If not, the error message might be deceiving.

third_party/e2e/engine.cc
```diff
@@ -0,0 +1,90 @@
...
+    ScsiToNvmeResponse resp = {.return_code = 0x40, .alloc_len = 0};
```
oicirtap, line 38: Can we add a named constant for `0x40 `? To make clear what error code is being returned.

third_party/e2e/engine.cc
```diff
@@ -0,0 +1,90 @@
...
+  translator::Span<const translator::NvmeCmdWrapper> nvme_wrappers =
+      translation.GetNvmeWrappers();
+  nvme::GenericQueueEntryCpl cpl_buf[nvme_wrappers.size()] = {};
+  for (uint32_t i = 0; i < nvme_wrappers.size(); ++i) {
+    NvmeCommand tmp_cmd;
+    NvmeCompletion tmp_cpl = {};
+    memcpy(&tmp_cmd, &nvme_wrappers[i].cmd, sizeof(tmp_cmd));
+    static_assert(sizeof(tmp_cmd) == sizeof(nvme_wrappers[i].cmd));
+    void* buffer = reinterpret_cast<void*>(nvme_wrappers[i].cmd.dptr.prp.prp1);
+    unsigned bufflen = 4096;
```
oicirtap, line 60: if this is a constant, please name it accordingly.

oicirtap, line 60: Is there a reason we don't want to specify the type here? To use int or long or something more specific

third_party/e2e/engine.cc
```diff
@@ -0,0 +1,90 @@
...
+    if (nvme_wrappers[i].is_admin) {
+      submit_admin_command(&tmp_cmd, buffer, bufflen, &tmp_cpl, 60);
```
oicirtap, line 62: Same here, please add a name for `60`. I know there's a lot of magic constants by the nature of the project, but adding names does make things more readable.

third_party/e2e/nvme_driver.c
```diff
@@ -0,0 +1,178 @@
...
+  u8 flags;
+  u16 status;
+  void* ctrl;
+};
+
+static inline struct nvme_request* nvme_req(struct request* req) {
+  return blk_mq_rq_to_pdu(req);
+}
+
+static void submit_req_done(struct request* rq) {
```
oicirtap, line 40: I'd use `request` here. In general, avoid shortening variable names by deleting letters.

https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

third_party/e2e/nvme_driver.c
```diff
@@ -0,0 +1,178 @@
...
+  struct nvme_request* nvme_rq = nvme_req(rq);
+
+  if (rq) {
+    blk_mq_free_request(rq);
+  }
+}
+
+struct request* nvme_alloc_request(struct request_queue* q,
+                                   struct nvme_command* cmd) {
```
oicirtap, line 49: here, names like `cmd` and `op` are in general ok because they are widely used/understood.

third_party/e2e/nvme_driver.c
```diff
@@ -0,0 +1,178 @@
...
+int nvme_submit_user_cmd(struct gendisk* disk, struct request_queue* q,
+                         struct nvme_command* cmd, void* buffer,
+                         unsigned bufflen, struct NvmeCompletion* cpl,
+                         unsigned timeout) {
+  struct request* req;
+  struct bio* bio = NULL;
+  int ret;
+
+  if (!q) {
+    printk("Request q is nullptr");
```
oicirtap, line 75: queue

third_party/e2e/engine.cc
oicirtap: If you have time to do so, do move out of third party anything that was developed by the team.

third_party/e2e/engine.cc
```diff
@@ -0,0 +1,91 @@
...
+    ScsiToNvmeResponse resp = {
+        .return_code = static_cast<uint8_t>(scsi::Status::kTaskAborted),
+        .alloc_len = 0};
+    return resp;
+  }
+
+  // Grab NVMe cmds and call NVMe interface
+  translator::Span<const translator::NvmeCmdWrapper> nvme_wrappers =
+      translation.GetNvmeWrappers();
+  nvme::GenericQueueEntryCpl cpl_buf[nvme_wrappers.size()] = {};
```
oicirtap, line 54: cpl?

third_party/e2e/engine.cc
Ralith: I think this does actually belong in `third_party` as it's logically part of the kernel modules, which themselves must be in third_party due to different licensing requirements than the rest of the project.

third_party/e2e/engine.cc
Ralith: Looks like a shorthand for 'completion' with widespread use in the project.

third_party/e2e/engine.cc
eyhhuang: Oops I forgot to link the PR that solved the issue, but there we have a `kPageSize = 4096` constant https://github.com/googleinterns/scsi2nvme/pull/199

Ralith:
I'll squash this manually to deal with the CLA issue once it's ready.

PR #159, opened 2020-07-28
Refactor inquiry and read capacity to use SafePointerCast

harloff-g:
Uses casting helper defined in #156 to reduce stack size.

This is necessary due to kernel stack limitations


lib/translator/inquiry.cc
```diff
@@ -342,17 +342,19 @@ StatusCode InquiryToScsi(Span<const uint8_t> raw_scsi, Span<uint8_t> buffer,
...
-  nvme::IdentifyNamespace identify_ns = {};
-  if (!ReadValue(ns_span, identify_ns)) {
-    DebugLog("Malformed IdentifyNamespace Data");
-    return StatusCode::kInvalidInput;
-  };
+  const nvme::IdentifyNamespace* identify_ns =
+      SafePointerCastRead<nvme::IdentifyNamespace>(ns_span);
+  if (identify_ns == nullptr) {
+    DebugLog("Identify namespace structure failed to cast");
+    return StatusCode::kFailure;
```
eyhhuang, line 356: I noticed that we don't have a test for this case. Do you think it's worth adding one?

harloff-g: Good point, I went ahead & added some basic tests to ensure the nullptr case is caught by the functions

Issue #155, opened 2020-07-24
Ensure Endian correctness on merged code


eyhhuang:
Ensure that translation libraries already merged to main account for endianess of host/network and endianness of scsi/nvme

PR #160, opened 2020-07-29
Correct endianess in Inquiry and Read Capacity 10

eyhhuang:
Closes https://github.com/googleinterns/scsi2nvme/issues/155


lib/translator/request_sense.cc
```diff
@@ -70,7 +76,8 @@ StatusCode RequestSenseToNvme(Span<const uint8_t> scsi_cmd,
     DebugLog("Malformed RequestSense Command");
     return StatusCode::kInvalidInput;
   }
-  allocation_length = request_sense_cmd.allocation_length;
+  allocation_length =
+      static_cast<uint32_t>(ntohs(request_sense_cmd.allocation_length));
```
harloff-g, line 80: Isn't request sense allocation length a single byte?

eyhhuang: Good catch!

test/translator/inquiry_test.cc
```diff
@@ -14,6 +15,8 @@
 
 #include "lib/translator/inquiry.h"
 
+#include <netinet/in.h>
+
```
JustinAWei, line 19: do we need to check kernel here?

test/translator/read_capacity_10_test.cc
```diff
@@ -14,6 +14,8 @@
 
 #include "lib/translator/read_capacity_10.h"
 
+#include <netinet/in.h>
```
JustinAWei, line 17: do we need to check kernel here?

test/translator/read_capacity_10_test.cc
```diff
@@ -164,10 +162,9 @@ TEST_F(ReadCapacity10Test, BlocklengthLimit) {
             translator::ReadCapacity10ToScsi(buffer_, nvme_cmds_[0]));
   scsi::ReadCapacity10Data result = {};
   ASSERT_TRUE(translator::ReadValue(buffer_, result));
-  EXPECT_EQ(result.returned_logical_block_address,
-            identify_ns_.nsze > 0xffffffff ? 0xffffffff : identify_ns_.nsze);
+  EXPECT_EQ(result.returned_logical_block_address, 0);
   EXPECT_EQ(result.block_length,
-            identify_ns_.lbaf[identify_ns_.flbas.format].lbads);
+            __bswap_32(identify_ns_.lbaf[identify_ns_.flbas.format].lbads));
```
JustinAWei, line 167: is this compatible w both le, be?

eyhhuang: this currently is. but i'll change to using htonl and htoll for more clarify

test/translator/inquiry_test.cc
eyhhuang: I think it's reasonable to expect that the tests will be run in the userspace

PR #161, opened 2020-07-30
Translate synchronize cache 10

eyhhuang:
Closes https://github.com/googleinterns/scsi2nvme/issues/131


lib/translator/translation.cc
```diff
@@ -159,6 +162,10 @@ ApiStatus Translation::Complete(Span<const nvme::GenericQueueEntryCpl> cpl_data,
       pipeline_status_ = ReadToScsi(buffer, nvme_cmds_[0], kLbaSize);
       ret = ApiStatus::kSuccess;
       break;
+    case scsi::OpCode::kSync10:
```
JustinAWei, line 165: maybe put a todo if we need to handle something in the gen translator

third_party/spdk/BUILD
```diff
@@ -5,5 +5,6 @@ exports_files(["LICENSE"])
 cc_library(
   name = "nvme_lib",
   hdrs = ["nvme.h"],
+  deps = ["@com_google_absl//absl/base"],
```
JustinAWei, line 8: intentional?

eyhhuang: yeah, nvme needs the absl dependency for the ABSL_ATTRIBUTE_PACKED annotation. It seems like before the nvme lib was always linked with the scsi lib so it somehow got the absl dependency from the scsi dependency chain? But when you try to just use the nvme lib without the scsi lib, then it breaks.

lib/translator/translation.cc
eyhhuang: all the necessary translations should be handled by this https://github.com/googleinterns/scsi2nvme/pull/150 so once that is merged then we're all good.

lib/translator/BUILD
```diff
@@ -19,6 +19,7 @@ cc_library(
     ":read_lib",
     ":read_capacity_10_lib",
     ":request_sense_lib",
+    ":synchronize_cache_10_lib",
```
toadchild, line 22: Please consider naming it synchronize_cache rather than synchronize_cache_10 for consistency with other modules.

Issue #162, opened 2020-07-31
Create NVMe Command Wrapper


eyhhuang:
Create a wrapper object that contains a `GenericQueueEntryCmd nvme_cmd` and a `boolean is_admin`

PR #164, opened 2020-07-31
Nvme command wrapper

eyhhuang:
Creates a 
``` 
struct NvmeCmdWrapper {
  nvme::GenericQueueEntryCmd cmd;
  bool is_admin;
}; 
```
to report to the NVMe driver whether the command created is an admin command.

This also fixes a bug in Inquiry where `InquiryToScsi` and `InquiryToNvme` had a discrepancy as to whether `identify_ns` or `identify_ctrl` corresponded to `nvme_cmds_[0]` or `nvme_cmds_[1]`
Now, `identify_ns = nvme_cmds_[0]` and is the first parameter passed into the `InquiryTo[Scsi/Nvme]`
and `identify_ctlr = nvme_cmds_[1]` and is the second parameter passed into the `InquiryTo[Scsi/Nvme]`

Closes https://github.com/googleinterns/scsi2nvme/issues/162


lib/translator/inquiry.cc
```diff
@@ -335,36 +335,37 @@ StatusCode InquiryToNvme(Span<const uint8_t> raw_scsi,
 
 // Main logic engine for the Inquiry command
 StatusCode InquiryToScsi(Span<const uint8_t> raw_scsi, Span<uint8_t> buffer,
-                         Span<const nvme::GenericQueueEntryCmd> nvme_cmds) {
```
eyhhuang, line 338: Passing `identify_ns` and `identify_ctrl` instead of the span to be consistent with the param list in `InquiryToNvme`

This also addresses the bug where `InquiryToNvme` had `identify_ns = nvme_cmds_[0]` but `InquiryToScsi` had `identify_ns = nvme_cmds_[1]`

harloff-g:
Individual command logic also has to set is_admin true or false. As such, we should probably refactor command arguments in ToNvme functions to NvmeCmdWrapper& rather than GenericQueueEntryCmd&, and set is_admin accordingly

test/translator/inquiry_test.cc
```diff
@@ -77,23 +77,25 @@ TEST_F(InquiryTest, InquiryToNvme) {
   translator::Allocation allocations[2] = {{}};
 
   translator::StatusCode status = translator::InquiryToNvme(
-      scsi_cmd_, nvme_cmds_[1], nvme_cmds_[0], alloc_len, nsid, allocations);
+      scsi_cmd_, nvme_cmds_[0], nvme_cmds_[1], alloc_len, nsid, allocations);
```
eyhhuang, line 80: Change in ordering is deliberate, see PR description https://github.com/googleinterns/scsi2nvme/pull/164#issue-460531001

eyhhuang:
> Individual command logic also has to set is_admin true or false. As such, we should probably refactor command arguments in ToNvme functions to NvmeCmdWrapper& rather than GenericQueueEntryCmd&, and set is_admin accordingly

Any reason individual command logic should set is_admin instead of it being set in translation.cc?

harloff-g:
> > Individual command logic also has to set is_admin true or false. As such, we should probably refactor command arguments in ToNvme functions to NvmeCmdWrapper& rather than GenericQueueEntryCmd&, and set is_admin accordingly
> 
> Any reason individual command logic should set is_admin instead of it being set in translation.cc?

I think it makes it easier to unit test & is more consistent with the translation.cc not having to implement command specific logic. Also commands with variable command ratio such as mode sense will need to implement is_admin inside the ToNvme function & having all interfaces do this makes it more consistent

JustinAWei:
We've been meaning to create some structure like below but haven't gotten around to it. What do y'all think?

```
union {
    AdminCmd
    Cmd
}

**Comment by eyhhuang** [comment]
> We've been meaning to create some structure like below but haven't gotten around to it. What do y'all think?
> 
> ```
> union {
>     AdminCmd
>     Cmd
> }
> ```

That's an interesting suggestion. Right now both admin and non-admin commands share the `nvme::GenericQueueEntryCmd` struct and we can only differentiate them based on the OpCode set. If we were to make the `AdminCmd` and `Cmd` structs, would they just be a wrapper around `nvme::GenericQueueEntryCmd` and also have a fixed boolean `is_admin` value?

File: test/translator/inquiry_test.cc [hunk 3]
```diff
@@ -76,24 +76,29 @@ TEST_F(InquiryTest, InquiryToNvme) {
   uint32_t alloc_len;
   translator::Allocation allocations[2] = {{}};
 
-  translator::StatusCode status = translator::InquiryToNvme(
-      scsi_cmd_, nvme_cmds_[1], nvme_cmds_[0], alloc_len, nsid, allocations);
+  translator::StatusCode status =
+      translator::InquiryToNvme(scsi_cmd_, nvme_wrappers_[0], nvme_wrappers_[1],
```
eyhhuang, line 80: Change in ordering is deliberate, see PR description https://github.com/googleinterns/scsi2nvme/pull/164#issue-460531001

JustinAWei:
> If we were to make the

oh yep, looks like the union idea was meant for opcodes. nevermind

PR #166, opened 2020-08-03
Write value with num_bytes

eyhhuang:
Adds a `WriteValue()` templated function with a `num_bytes` parameter to specify how many bytes to write to the buffer instead of writing the whole object. This is needed in cases such as `InquiryData`, which is 96 bytes but only 36 of which is to be written to the buffer.

This change also ensures every invocation of `WriteValue()` has a check for success/failure to ease with debugging

Closes https://github.com/googleinterns/scsi2nvme/issues/165


eyhhuang:
> PR looks good but a general comment about the composition - the part where you added status code to Inquiry functions should have been a separate PR as it isn't related to changing the WriteValue function.

That's a great point! I'll definitely keep that in mind for future changes!

lib/translator/common.h
```diff
@@ -130,13 +130,18 @@ bool ReadValue(Span<const uint8_t> data, T& out) {
 }
 
 template <typename T>
-bool WriteValue(const T& data, Span<uint8_t> out) {
+bool WriteValue(const T& data, Span<uint8_t> out, size_t num_bytes) {
```
Ralith, line 133: This is going to cause a lot of churn, and additionally undermines safety guarantees as nothing stops `num_bytes` from being larger than `sizeof(T)`. Could it be introduced as an overload for the presumably rare cases that need it, so the rest can use the simpler and shorter function? Or better yet, could the case that needs to write only a prefix of a struct just use a dedicated struct definition containing only that prefix?

eyhhuang: That's a very valid concern of `num_bytes` exceeding `sizeof(T)`. I think it's worth adding a check to error if we continue with this implementation 

With respect to the overload, this function currently is an overload of the other `WriteValue()` function that used in most cases (https://github.com/googleinterns/scsi2nvme/pull/166/files#diff-e8eb0eb4e8ba5e2a5b4c9411b3673f7dR141), however, I think just having one function where `num_bytes` has a default value would be clearer. 

I think the dedicated struct idea could also be a good option, especially since that right now we are only aware of one such use case. However, if we discover more cases that require writing a prefix of the struct, then having this `WriteValue` function might make it cleaner.

Ralith: Ah, somehow I missed the overload below. Thanks!

Issue #169, opened 2020-08-04
Fix read capacity block size


harloff-g:
Read capacity does not bitshift the block size & reports it in little-endian

PR #170, opened 2020-08-04
Read capacity endian

eyhhuang:
Closes https://github.com/googleinterns/scsi2nvme/issues/169


lib/translator/read_capacity_10.cc
```diff
@@ -75,7 +75,7 @@ StatusCode ReadCapacity10ToScsi(
           htonl(ltohll(identify_ns->nsze) > 0xffffffff
                     ? 0xffffffff
                     : static_cast<uint32_t>(ltohll(identify_ns->nsze))),
-      .block_length = htoll(static_cast<uint32_t>(
+      .block_length = htonl(static_cast<uint32_t>(
```
harloff-g, line 78: This result should be bit-shifted (1 << identify_ns->lbaf[identify_ns->flbas.format].lbads). SCSI expected the result in bytes, NVMe uses a power of 2 identifier

PR #173, opened 2020-08-05
Set alloc_len in ReadToNvme and VerifyReportSupportedOpCodes

eyhhuang:
These two translation libraries were previously not setting the allocation length. The allocation length is necessary to set the size of the `data_in` buffer that the translated response will be written to.

Closes https://github.com/googleinterns/scsi2nvme/issues/172


basimsahaf:
Is this change required for `Write` command as well?

eyhhuang:
> Is this change required for `Write` command as well?

Setting the alloc_len is required every time we need to write command specific data to the scsi `data_in` buffer. I took a quick look at the Write PRs and it seems like Write does not write to the buffer. If the response of Write is just status codes, then that is handled separately. So I don't think this change is needed for Write.

PR #186, opened 2020-08-10
Adding buffer to the read command

basimsahaf:
This PR passes the incoming SCSI buffer for read command. It also changes the read tests so that a buffer of the appropriate length can be passed to the various Read commands

TODOs:
- [ ] Merge with `nvme-engine` branch to pass the SCSI buffer and test. I will be doing that on a different branch because `nvme-engine` has not been merged yet

### DIFF
--- a/lib/translator/read.cc
+++ b/lib/translator/read.cc
@@ -73,7 +73,7 @@ uint16_t GetTransferLengthPages(uint16_t transfer_length, uint32_t page_size,
 StatusCode LegacyRead(nvme::GenericQueueEntryCmd& nvme_cmd,
                       Allocation& allocation, uint32_t nsid,
                       uint16_t transfer_length, uint32_t page_size,
-                      uint32_t lba_size, uint32_t& alloc_len) {
+                      uint32_t lba_size, uint32_t& alloc_len, Span<const uint8_t> buffer_in) {
   StatusCode status = allocation.SetPages(
       GetTransferLengthPages(transfer_length, page_size, lba_size), 0);
   if (status != StatusCode::kSuccess) {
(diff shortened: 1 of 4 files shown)


lib/translator/read.cc
```diff
@@ -85,18 +86,24 @@ StatusCode LegacyRead(nvme::GenericQueueEntryCmd& nvme_cmd,
...
-
   alloc_len = transfer_length * lba_size;
 
+  // checks beforehand if enough space has been allocated for the buffer
+  if (buffer_in.size() < alloc_len) {
+    DebugLog("Not enough memory allocated for Read buffer");
+    return StatusCode::kFailure;
+  }
+
+  nvme_cmd.dptr.prp.prp1 = reinterpret_cast<uint64_t>(buffer_in.data());
```
basimsahaf, line 97: This is where the prp pointer is assigned the address of the memory to where the buffer is pointing towards. I switched the logic to check the size of the buffer beforehand rather than checking it in the `ReadToScsi` function as that is more efficient since no call will be made to the device in case the buffer size is incorrect.

lib/translator/read.cc
```diff
@@ -250,11 +260,6 @@ StatusCode ReadToScsi(Span<uint8_t> buffer,
 
   uint32_t bytes_transferred = transfer_length * lba_size;
 
-  if (buffer.size() < bytes_transferred) {
-    DebugLog("Insufficient buffer size");
-    return StatusCode::kFailure;
-  }
-
```
basimsahaf, line 257: https://github.com/googleinterns/scsi2nvme/pull/186/files#r468090300

eyhhuang:
Does this mean that we can get rid of the `ReadToScsi` command?

basimsahaf:
Refer to #188

basimsahaf:
> Does this mean that we can get rid of the `ReadToScsi` command?

Yes we can remove that function now! Check #188

PR #188, opened 2020-08-10
Adding SCSI buffer to Read Command

basimsahaf:
This PR passes the incoming SCSI buffer for read command. It also changes the read tests so that a buffer of the appropriate length can be passed to the various Read commands

TODOs:
- [x] Merge with `nvme-engine` branch to pass the SCSI buffer and test. I will be doing that on a different branch because `nvme-engine` has not been merged yet

### DIFF
--- a/lib/translator/read.cc
+++ b/lib/translator/read.cc
@@ -87,7 +87,6 @@ StatusCode LegacyRead(nvme::GenericQueueEntryCmd& nvme_cmd,
       .nsid = nsid};
 
   alloc_len = transfer_length * lba_size;
-  buffer_in = Span<uint8_t>(buffer_in.data, alloc_len);
 
   // // checks beforehand if enough space has been allocated for the buffer
   // if (buffer_in.size() < alloc_len) {
--- a/lib/translator/read.cc
+++ b/lib/translator/read.cc
@@ -253,19 +253,4 @@ StatusCode Read16ToNvme(Span<const uint8_t> scsi_cmd,
   return StatusCode::kSuccess;
 }
 
-StatusCode ReadToScsi(Span<uint8_t> buffer,
-                      const nvme::GenericQueueEntryCmd& nvme_cmd,
-                      uint32_t lba_size) {
-  // Get NVMe read data
-  uint64_t data_ptr = nvme_cmd.dptr.prp.prp1;
-  // cdw 12 nlb bits 15:00 (zero based field)
-  uint16_t transfer_length = static_cast<uint16_t>(ltohl(nvme_cmd.cdw[2]) + 1);
-
-  uint32_t bytes_transferred = transfer_length * lba_size;
-
-  memcpy(buffer.data(), reinterpret_cast<uint8_t*>(data_ptr),
-         bytes_transferred);
-  return StatusCode::kSuccess;
-}
-
 }  // namespace translator
--- a/lib/translator/read.cc
+++ b/lib/translator/read.cc
@@ -86,7 +86,7 @@ StatusCode LegacyRead(nvme::GenericQueueEntryCmd& nvme_cmd,
       .psdt = 0,  // PRPs are used for data transfer
       .nsid = nsid};
 
-  alloc_len = transfer_length * lba_size;
+  alloc_len = transfer_length * page_size;
 
   // // checks beforehand if enough space has been allocated for the buffer
   // if (buffer_in.size() < alloc_len) {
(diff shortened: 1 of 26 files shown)


lib/translator/read.cc
```diff
@@ -86,18 +86,25 @@ StatusCode LegacyRead(nvme::GenericQueueEntryCmd& nvme_cmd,
...
   alloc_len = transfer_length * lba_size;
 
+  // checks beforehand if enough space has been allocated for the buffer
+  if (buffer_in.size() < alloc_len) {
+    DebugLog("Not enough memory allocated for Read buffer");
+    DebugLog("Memory required: %d", alloc_len);
+    return StatusCode::kFailure;
+  }
+
+  nvme_cmd.dptr.prp.prp1 = reinterpret_cast<uint64_t>(buffer_in.data());
```
basimsahaf, line 98: This is where the prp pointer is assigned the address of the memory to where the buffer is pointing towards. I switched the logic to check the size of the buffer beforehand rather than checking it in the ReadToScsi function as that is more efficient since no call will be made to the device in case the buffer size is incorrect.

lib/translator/read.h
```diff
@@ -35,22 +35,22 @@ namespace translator {
...
-                        uint32_t& alloc_len);
+                        uint32_t& alloc_len, Span<const uint8_t> buffer_in);
 
 StatusCode Read16ToNvme(Span<const uint8_t> scsi_cmd,
                         NvmeCmdWrapper& nvme_wrapper, Allocation& allocation,
                         uint32_t nsid, uint32_t page_size, uint32_t lba_size,
-                        uint32_t& alloc_len);
+                        uint32_t& alloc_len, Span<const uint8_t> buffer_in);
 
 // Common to Read(6), Read(10), Read(12), and Read(16)
```
eyhhuang, line 55: I think `ReadToScsi` still needs to be removed from the header file

lib/translator/report_luns.cc
```diff
@@ -91,17 +90,14 @@ StatusCode ReportLunsToScsi(const nvme::GenericQueueEntryCmd& identify_cmd,
...
   uint32_t lun_count = GetNsListLength(*ns_list);
-  // TODO: TEMP TESTING.
-  // uint32_t lun_count = 1;
 
   // Define scsi response structure and determine attributes
   uint32_t allocated_list_bytes =
       buffer.size() - sizeof(scsi::ReportLunsParamData);
   scsi::ReportLunsParamData rlpd = {};
   rlpd.list_byte_length = htonl(lun_count * sizeof(scsi::LunAddress));
-  if (ntohl(rlpd.list_byte_length) > allocated_list_bytes) {
```
eyhhuang, line 104: we still need to call ntohl here because in the line above we made rlpd.list_byte_length big endian.

lib/translator/translation.cc
```diff
@@ -99,25 +99,25 @@ BeginResponse Translation::Begin(Span<const uint8_t> scsi_cmd,
     case scsi::OpCode::kRead6:
       pipeline_status_ =
           Read6ToNvme(scsi_cmd_no_op, nvme_wrappers_[0], allocations_[0], nsid,
-                      kPageSize, kLbaSize, response.alloc_len);
+                      kPageSize, kLbaSize, response.alloc_len, buffer_out);
```
eyhhuang, line 102: a little confusing that the parameter passed in here is called `buffer_out` but the parameter for the read_functions is called `buffer_in`. It might be helpful to add a comment somewhere to clarify. Or, if there's already a comment somewhere, then you can ignore this :)

test/translator/read_test.cc
```diff
@@ -100,6 +107,7 @@ TEST_F(ReadTest, Read6ToNvmeShouldRead256BlocksForZeroTransferLen) {
   uint8_t network_endian_lba_1 = 0x1a;
   uint16_t network_endian_lba_2 = htons(0x2b3c);
   uint32_t cdw10 = 0x1a2b3c;
+  uint16_t transfer_len = 256;
```
eyhhuang, line 110: Can you add a comment here that technically the transfer_length is set to zero when building the SCSI command but it "means" 256

eyhhuang, line 110: Also, what are your thoughts on just having a `kBufferIn` global variable that has a size big enough for all the tests. I think that might help make things more readable. I'm also fine just leaving it as is.

eyhhuang:
Closed and replaced with https://github.com/googleinterns/scsi2nvme/pull/197

PR #189, opened 2020-08-11
Check for ApiStatus::kFailure

eyhhuang:


third_party/scsi_mock/engine.cc
```diff
@@ -71,6 +78,13 @@ ScsiToNvmeResponse ScsiToNvme(unsigned char* cmd_buf, unsigned short cmd_len,
   translator::CompleteResponse cpl_resp =
       translation.Complete(nvme_cpl, buffer_in, sense_buffer);
 
+  if (cpl_resp.status == ApiStatus::kFailure) {
+    Print("Incorrect usage of Translation Library API");
+    ScsiToNvmeResponse resp = {.return_code = 0x40, .alloc_len = 0};
+    translation.AbortPipeline();
```
Ralith, line 84: Isn't this already called by `Complete`?

third_party/scsi_mock/engine.cc
```diff
@@ -33,6 +33,13 @@ ScsiToNvmeResponse ScsiToNvme(unsigned char* cmd_buf, unsigned short cmd_len,
   translator::BeginResponse begin_resp =
       translation.Begin(scsi_cmd, buffer, lun);
 
+  if (begin_resp.status == ApiStatus::kFailure) {
+    Print("Incorrect usage of Translation Library API");
+    ScsiToNvmeResponse resp = {.return_code = 0x40, .alloc_len = 0};
+    translation.AbortPipeline();
```
Ralith, line 39: Do we actually leave any resources allocated in the case of an early failure?

eyhhuang: Good catch! In Translation::Begin, the only case of a `ApiStatus::kFailure` occurs before any resource allocation.

third_party/scsi_mock/engine.cc
eyhhuang: When `Complete` returns an `ApiStatus::kFailure`, it doesn't call `AbortPipeline`. I'm still investigating whether it is necessary to call `AbortPipeline`. In any case, the user (`engine.cc`) probably shouldn't be responsible for calling `AportPipeline`

Ralith: Ah, I see e.g. https://github.com/googleinterns/scsi2nvme/blob/master/lib/translator/translation.cc#L153-L160 could cause a failure return with allocated resources. I agree this should probably be changed to clean up automatically.

PR #192, opened 2020-08-17
Read directly from buffer

eyhhuang:
Proof of concept, do not merge


lib/translator/read.cc
```diff
@@ -169,7 +169,7 @@ StatusCode Read6ToNvme(Span<const uint8_t> scsi_cmd,
 StatusCode Read10ToNvme(Span<const uint8_t> scsi_cmd,
                         NvmeCmdWrapper& nvme_wrapper, Allocation& allocation,
                         uint32_t nsid, uint32_t page_size, uint32_t lba_size,
-                        uint32_t& alloc_len) {
+                        uint32_t& alloc_len, uint64_t ptr) {
```
Ralith, line 172: `alloc_len`, being an output argument, should remain at the end of the argument list.

Issue #196, opened 2020-08-21
Handle multiple pages in NVMe driver


eyhhuang:
Right now the `bufferlen` in `engine.cc` is hardcoded to 4096, which means we can only send a page of data. We should change this to reflect the actual bytes of data being transferred

Ralith:
Fixed by #199.

PR #199, opened 2020-08-24
Nvme multiple pages

eyhhuang:
Currently,  the `bufflen` in `engine.cc` is hardcoded to 4096, which means we can only handle one page of data. This change adds a field `buffer_len` to the `NvmeCmdWrapper` object. Each command translation sets `buffer_len` to the amount of memory that is allocated for the PRP pointer. This also refactors `AllocPages` to take in the `page_size` as a parameter.

This PR also handles some minor refactors such as changing the order or the parameter list to maintain consistency.
Closes https://github.com/googleinterns/scsi2nvme/issues/196


lib/translator/inquiry.cc
```diff
@@ -567,8 +567,8 @@ StatusCode TranslateLogicalBlockProvisioningVpd(
 StatusCode InquiryToNvme(Span<const uint8_t> raw_scsi,
                          NvmeCmdWrapper& identify_ns_wrapper,
                          NvmeCmdWrapper& identify_ctrl_wrapper,
-                         uint32_t& alloc_len, uint32_t nsid,
-                         Span<Allocation> allocations) {
+                         uint32_t page_size, uint32_t nsid,
+                         Span<Allocation> allocations, uint32_t& alloc_len) {
```
eyhhuang, line 571: Changing param order so output param is at the end, which is also consistent with other command translation libraries

test/translator/translation_test.cc
```diff
@@ -49,7 +50,7 @@ TEST(Translation, ShouldFailInvalidPipeline) {
       translation.Complete(cpl_data, buffer_in, sense_buffer);
   EXPECT_EQ(translator::ApiStatus::kFailure, status.status);
 }
-
+/**
```
toadchild, line 53: Why is this test being commented out?

eyhhuang: Good catch! This was an accident.

